PR #2, opened 2016-07-16
Maven and refactoring

nateschickler0:
Library .jar files should not be present in the repo, used Maven to offload repos and use external dependencies. Refactored packages to improve readability and structure. No changes were made to the functionality of the code.


Grover-c13:
Thanks for cleaning up :), may  change naming scheme in future again though but i agree this is better for packages.

PR #8, opened 2016-07-17
refractor code using lombok

vmarchaud:
- refractored getter/setter with lombok (lombok generate them at compile time)
- add a build rule to bundle all dependency into a single jar
- some other refractor

i have compulsively sometime put the first { on the same ligne of the function, sorry to mess up the identation ^^

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,8 @@ local.properties
 .settings/
 .loadpath
 .recommenders
+.classpath
+
 
 # Eclipse Core
 .project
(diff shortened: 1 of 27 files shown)


vmarchaud:
I forgot that i used java 8 stream api to rewrite [some list](https://github.com/vmarchaud/PokeGOAPI-Java/blob/refractor/src/com/pokegoapi/api/inventory/PokeBank.java#L29) [related stuff](https://github.com/vmarchaud/PokeGOAPI-Java/blob/refractor/src/com/pokegoapi/api/inventory/PokeBank.java#L33)

Issue #23, opened 2016-07-18
com.pokegoapi.main.Request 404 Not found


vmarchaud:
Has you can read, someone deleted it, its gonna be easier if someone can re-push it again ;)
PS : gonna submit a PR to add how to build/install the api to the readme, might push it with it.

langerhans:
A new mechanism for this was pushed but it seems the commit was incomplete. So not sure if it's supposed to stay. If you revert your branch by two commits you'll have a working base again.

vmarchaud:
Yeah i was trying to understand why the master branch is used to this kind of rewrite.

langerhans:
Agree, would be a good idea to at least have a master/develop workflow or just do **everything** in PRs.

vmarchaud:
I dont even know if its usefull to finish the work and submit the PR or the author has already finish but not pushed ...

jabbink:
Will revert `master` to a working state and move those 2 commits to a branch so @Grover-c13 can fix it tomorrow.

jabbink:
Fixed

PR #26, opened 2016-07-18
[WIP] Fix temp remove server request

jabbink:
@Grover-c13 see #23 

Had to revert this and store the commits here. Seems like you missed some files before you went offline.


Grover-c13:
ma bad

PR #31, opened 2016-07-19
Add in commented out code in Map

Aphoh:


jabbink:
Closing in favor of #32 as the `missedCells` property is not working (server always returns all cells, but sometimes with an empty response).

PR #66, opened 2016-07-20
Map API, added getMap().getCatchablePokemon(), added CatchablePokemon class, (can do catchablePokemon.catch(), or catchablePokemon.catch(Pokeballtype)

Grover-c13:
Map API, added getMap().getCatchablePokemon(), added CatchablePokemon class, (can do catchablePokemon.catch(), or catchablePokemon.catch(Pokeballtype)


Grover-c13:
Tested and working

PR #76, opened 2016-07-20
add the deletion of an item + some rewrite of exception handling

vmarchaud:
need testing before merging, will wait that server come back.


FabianTerhorst:
server should be up now

vmarchaud:
not enough stable to test it correctly :(

jabbink:
tested and works

PR #79, opened 2016-07-20
Pulled down AeonLucid's proto files today and built into a jar

akirarat:
I checked it into the repo until there's a better way to include this into the dependencies.

Added Slf4j for logging.
Added one more example file PokeBankList which will just list the pokemons you have.
Added a UserReader that allows you to input tokens and login to POGO without hard coding.


FabianTerhorst:
Are you sure that the user reader is in the scope of this Library?

akirarat:
I'll take it out.

vmarchaud:
We do not need any hard jar dependency with protos since we are generating them too.
To update to the last commit, just update the submodule :)

PR #82, opened 2016-07-21
PTCLogin refactored without changing the logic

alessiop86:
I have reduced some duplication and refactored the methods in order to made evident what's happening in the oauth flow (credentials sent --> authentication ticket issued -->new oauth token generated)

I have not changed the logic except for the exception cause now thrown inside the login exceptions


FabianTerhorst:
Are you sure that that api doesn´t support post parameters as well? For example: 

```
RequestBody formBody = new FormBody.Builder()
                                        .add("lt", authorizeEntity.getLt())
                                        .add("execution", authorizeEntity.getExecution())
                                        .add("_eventId", "submit")
                                        .add("username", username)
                                        .add("password", password)
                                        .build();

                                Request interceptRedirect = new Request.Builder()
                                        .addHeader("User-Agent", "Niantic App")
                                        .url(LOGIN_URL)
                                        .post(formBody)
                                        .build();
```

is actually working.

alessiop86:
@FabianTerhorst I am not sure, I just refactored as it was. I believe we should try to replicate the behaviour of the official client, do you know if the mobile apps Are using a POST or a GET call to get the authentication tickets?

FabianTerhorst:
@alessiop86 The default behavior to submit passwords is a post call. It's not a security reason.

alessiop86:
Oh sorry, I replied in hurry this morning, just before going to work and I misunderstood you, I thought you were questioning whether the call was supposed to be a GET or POST.

Anyway, I introduced the `.post()` version as you suggested

Grover-c13:
Thank you for this, ill be refactoring login stuff tonight so will probably just do this manually instead of merging (<_< >_>).

PR #83, opened 2016-07-21
CatchablePokemon refactor

mjmjelde:
Moved encounter and catching code to CatchablePokemon.  Also implemented retrying of pokemon if pokemon escapes or throw misses if user wants.

### DIFF
--- a/src/main/java/com/pokegoapi/api/map/Pokemon/EncounterResult.java
+++ b/src/main/java/com/pokegoapi/api/map/Pokemon/EncounterResult.java
@@ -1,0 +1,23 @@
+package com.pokegoapi.api.map.Pokemon;
+
+import POGOProtos.Networking.Responses.EncounterResponseOuterClass.EncounterResponse;
+
+/**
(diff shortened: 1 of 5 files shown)


src/main/java/com/pokegoapi/api/map/Pokemon/EncounterResult.java
```diff
@@ -0,0 +1,23 @@
...
+import POGOProtos.Networking.Responses.EncounterResponseOuterClass.EncounterResponse;
+
+/**
+ * Created by mjmfighter on 7/20/2016.
+ */
+public class EncounterResult {
+
+    private EncounterResponse response;
+
+    public EncounterResult(EncounterResponse response){
```
FabianTerhorst, line 12: indent this class please

jabbink, line 12: Did it for him and merged it. (tested it as well, and works in my demo project)

PR #87, opened 2016-07-21
Added renaming to Pokemon.java

nateschickler0:
renamePokemon(String nickname) renames a pokemon to a given string.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -154,3 +154,4 @@ fabric.properties
 
 
 
+/src/resources/.gitignore
--- a/.gitignore
+++ b/.gitignore
@@ -154,4 +154,3 @@ fabric.properties
 
 
 
-/src/resources/.gitignore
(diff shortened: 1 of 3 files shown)


src/main/java/com/pokegoapi/api/pokemon/Pokemon.java
```diff
@@ -55,14 +57,31 @@ public Result transferPokemon() throws LoginFailedException, RemoteServerExcepti
...
+
+		ServerRequest serverRequest = new ServerRequest(RequestType.NICKNAME_POKEMON, reqMsg);
+		pgo.getRequestHandler().request(serverRequest);
+		pgo.getRequestHandler().sendServerRequests();
+
+		NicknamePokemonResponse response;
+		try {
+			response = NicknamePokemonResponse.parseFrom(serverRequest.getData());
+		} catch (InvalidProtocolBufferException e) {
+			return null;
```
jabbink: On all other places `throw new RemoteServerException(e);` is used so the error can be handled by the caller.

.gitignore
```diff
@@ -154,3 +154,4 @@ fabric.properties
 
 
 
+/src/resources/.gitignore
```
jabbink: ?

nateschickler0: Mistakenly added this, will remove it.

src/main/java/com/pokegoapi/api/pokemon/Pokemon.java
nateschickler0: The method throws RemoteServerException.
... `throws LoginFailedException, RemoteServerException {`

Grover-c13: the return null in the catch state should throw a RemoteServerException

nateschickler0: Got it. Adding now.

jabbink:
:+1:

PR #88, opened 2016-07-21
Add WildPokemon to Map.getCatchablePokemon, as they are catchable

jabbink:


jabbink:
Tested, and was able to immediately catch a `WildPokemon`

PR #90, opened 2016-07-21
Map API

Grover-c13:
map.getNearbyPokemon()
NearbyPokemon object


Grover-c13:
Tested & Working, also added optionalb useCache field to map object to make it always get a new MapObjects on request.

Aphoh:
Are you merging this?

PR #91, opened 2016-07-21
MapAPI

Grover-c13:


Grover-c13:
Map API - getSpawnPoints() getDecimatedSpawnPoints() point object

PR #108, opened 2016-07-22
Inventory refactor

mjmjelde:
Moved all the "inventory" based items out of the main api and into their own class.  This is just an inital setup, and can be changed to fit other needs.  Probably needs a bit more testing but works with my testing.  Supports updating of inventory and not just a full update.  May need to go to other classes and remove their interactions with the bags as when an update is called it could mess up.  That part is untested.


Grover-c13:
can you rebase and check it conforms to the style check, if all is good will merge

mjmjelde:
So many fixed and shit happened when I rebased this and fixed all the errors.  Its all fixed now I believe

PR #120, opened 2016-07-22
Updated read me with working java and gradle version

DeanIsseyegh:


jabbink:
Minimum JDK version is 1.7 as specified in `build.gradle`

Minimum Gradle version is `2.12` afaik.

Hindi:
You can't build with 1.7 because of the spell check plugin. It requires 1.8

jabbink:
I stand corrected.

Nice to be back at the beginning (I added Java 8 only code at the beginning of the week and people complained because this would break Android compatibility).

Now we're back at Java 8. smh

Hindi:
Maybe there is an alternative to checkStyle that run with jdk 1.7?
I don't think the plugin is worth losing Android compatibility.

langerhans:
Wait wait, there is a difference between source compatibility and some plugin's requirements. The source is still Java 6 compliant, therefore Android compatible. But the plugin itself needs a Java 8 VM to run on. So that is fine.

And yes, Gradle min is 2.12 as required by protobuf.

PR #126, opened 2016-07-22
Fix checkstyle for requestHandler

Hindi:
A samll checkstyle fix to prevent warnings.


jabbink:
Weird, my local checkstyle and travis said it was fine

https://github.com/Grover-c13/PokeGOAPI-Java/pull/122

https://travis-ci.org/Grover-c13/PokeGOAPI-Java/builds/146622925

Hindi:
Indeed it's weird.

capture

CheckStyle logs in French, but it says something like '&& should be on a new line'.

FabianTerhorst:
The code looks more readable with this change, even when its building without.

PR #127, opened 2016-07-22
Added Throws on GetPlayerProfile and PokemonGO contructor

pompobit:
- also fixed checkstyle issue


pompobit:
When servers have troubles, PokemonGO object often fails to contruct, so I think it's useful to throws exception to the user to handle them.
I also fixed checkstyle issue in this pr to build but I saw that there is another one just for this

jabbink:
I think it would be nicer to remove the "expensive" web call in the constructor and let users of the API call this initial profile call themselves.

PR #129, opened 2016-07-22
Fix catching of pokemon.

mjmjelde:
Also added updating of inventories to a few methods related to pokemon


mjmjelde:
There is another PR to fix this problem again.  Once its reviewed it will be merged and fixed agian

PR #134, opened 2016-07-22
Fixed pokemon showing up double in getCatchablePokemon()

Jari27:
This should fix it.


przerwap:
Would it not be simpler to just collect the pokemon into a Set

Jari27:
Probably not, since then you'd still have to check against their encounterId (`getEncounterId()`), since the actual objects aren't the same. Instead of breaking working things I decided to do it this way.

przerwap:
Equality of objects is determined though the equals() method. And Sets can't contain duplicates.

Jari27:
So you are suggesting changing the list to a collection, changing every piece of code that depends on that list being a list and overriding the `equals`method in the `CatchablePokemon`class?

Obviously that's a solution, but like I said before: I'm not breaking things that are working if I can help it.

przerwap:
Nah man. There are basically two thing you can do. Each includes defining that simple equals() and hashCode() method.

You can Just make a set .. which makes it so you don't have to do all that logic there like so:

``` Java
Set<CatchablePokemon> pokemon = new HashSet<>(catchablePokemons);

return new ArrayList<>(pokemon);
```

Or .. if you look a few lines down in the same class, you'll notice that there are Streams being used in the project.

``` Java
return RefStreams.concat(
        StreamSupport.stream(objects.getCatchablePokemons())
                .map(new Function<MapPokemon, CatchablePokemon>() {
                    @Override
                    public CatchablePokemon apply(MapPokemon mapPokemon) {
                        return new CatchablePokemon(api, mapPokemon);
                    }
                }),
        StreamSupport.stream(objects.getWildPokemons())
                .map(new Function<WildPokemonOuterClass.WildPokemon, CatchablePokemon>() {
                    @Override
                    public CatchablePokemon apply(WildPokemonOuterClass.WildPokemon mapPokemon) {
                        return new CatchablePokemon(api, mapPokemon);
                    }
                }))
        .distinct()
        .collect(Collectors.<CatchablePokemon>toList());
```

The second option is likely preferable I'd say, since it removes the need for those for loops mapping the Wild/Map Pokemon classes to CatchablePokemon. It's a shame we can't use actual lambda here, but oh well.

jabbink:
I think replacing `Collection<CatchablePokemon>` by `Set<CatchablePokemon>` and implementing the hashCode and equals method is quite elegant in this case.

Grover-c13:
I agree a set would be best, hashCode can just use the encounterid since it will be unique.

Jari27:
Done.

FabianTerhorst:
you have to rebase from the masterbranch und fix the codestyle inside the catchablepokemon.java

FabianTerhorst:
whats about: 

```
 if (o == this) {
    return true;
 } else if (o instanceof CatchablePokemon) {
        return getEncounterId() == ((CatchablePokemon) o).getEncounterId();
 }
return false
```

for https://github.com/Grover-c13/PokeGOAPI-Java/pull/134/files#diff-2abf18e8affc55dadaab37add72c02adR223
?

src/main/java/com/pokegoapi/api/map/Map.java
```diff
@@ -91,8 +95,8 @@ public void clearCache() {
 	 *
 	 * @return a List of CatchablePokemon at your current location
 	 */
-	public List<CatchablePokemon> getCatchablePokemon() throws LoginFailedException, RemoteServerException {
-		List<CatchablePokemon> catchablePokemons = new ArrayList<>();
+	public Set<CatchablePokemon> getCatchablePokemon() throws LoginFailedException, RemoteServerException {
```
przerwap: You can't change the response type here. That would affect anyone/anything that is currently using this method.

FabianTerhorst: He could return a list and transform the set into a arraylist with

```
new ArrayList<CatchablePokemon>(catchablePokemons);
```

Grover-c13:
Please rebase & conform to the style check (gradlbe build should report issues)

Jari27:
Uuuuhg it's messed up.

build.gradle
```diff
@@ -16,7 +16,6 @@ apply plugin: 'idea'
 apply plugin: 'java'
 apply plugin: 'maven'
 apply plugin: 'com.google.protobuf'
-apply plugin: 'checkstyle'
```
jabbink, line 19: That's a way to fix the checkstyle.

Jari27, line 19: I can't get checkstyle to work with my java version so I turned it off, but now it's committing the build.gradle file as well. Haven't worked with git that often :(

jabbink, line 19: `git commit -a` sucks

Jari27, line 19: Didn't even use that. I think it came from the rebasing. I only changed (and added and committed) Map.java and CatchablePokemon.java.

Jari27, line 19: And it seems there are errors in the code I pulled, which is weird.

Jari27:
I'm starting over.

PR #141, opened 2016-07-23
null pointer fix

mjmjelde:


mjmjelde:
Rebased with development brach.  Should be good to merge now

PR #153, opened 2016-07-23
Fix for issue 132 (pokemon double in list)

Jari27:
Retry. I didn't check style, I can't get it working. :/

### DIFF
--- a/src/main/java/com/pokegoapi/api/map/pokemon/CatchablePokemon.java
+++ b/src/main/java/com/pokegoapi/api/map/pokemon/CatchablePokemon.java
@@ -13,7 +13,7 @@
  *     along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-package nl.sadye.bot;
+package com.pokegoapi.api.map.pokemon;
 
 import POGOProtos.Enums.PokemonIdOuterClass;
 import POGOProtos.Inventory.Item.ItemIdOuterClass.ItemId;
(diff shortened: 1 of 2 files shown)


FabianTerhorst:
@Sadye have you tried to simplify the logic in the equals method to:

``` java
 if (o == this) {
    return true;
 } else if (o instanceof CatchablePokemon) {
        return getEncounterId() == ((CatchablePokemon) o).getEncounterId();
 }
return false
```

FabianTerhorst:
travis is failing, because of the missing HashMap import. 

``` java
import java.util.HashMap;
```

Jari27:
Agreed. Fixed both your suggestions.

FabianTerhorst:
You changed the package location with your latest commit and you have to indent the equals method again.

src/main/java/com/pokegoapi/api/map/pokemon/CatchablePokemon.java
```diff
@@ -235,12 +266,27 @@ public CatchResult catchPokemon(
...
 		api.getInventories().updateInventories();
 
 		return new CatchResult(response);
 	}
 
+	@Override
+	public boolean equals(Object o) {
+		if (o == this) {
+			return true;
+		} else if (o instanceof CatchablePokemon){
```
FabianTerhorst: add a space between ) and {

Jari27:
Thanks for coaching ;)

FabianTerhorst:
looks good to me, travis is building fine. :thumbsup:

Grover-c13:
-   public boolean equals(Object o) {
  -      if (o == this) {
  -          return true;
  -      } else if (o instanceof CatchablePokemon) {
  -          return this.getEncounterId() == ((CatchablePokemon) o)
  -                  .getEncounterId();
  -      }
  -      return false;
  -  }

single letter variables wont conform to the check style sorry, not sure why travis is passing

PR #154, opened 2016-07-23
ChatchResult.isFailed should only concern response issue or status.

Hindi:
In the previous version, isFailed can return true if response == null or if gestStatus() != CatchStatus.CATCH_SUCCESS.

It is a problem because the user may want to check why it failed, maybe it is because the inventory is full or because of a soft ban. To check this, the user will call getStatus(), but getStatus will throw a null reference exception if response == null.

I suggest that isFailed should only return true if response == null and the user handle the status value on his side.

### DIFF
--- a/src/main/java/com/pokegoapi/api/map/pokemon/CatchResult.java
+++ b/src/main/java/com/pokegoapi/api/map/pokemon/CatchResult.java
@@ -43,7 +43,7 @@ public class CatchResult {
 
 	public CatchStatus getStatus() {
 		return response.getStatus();
-	}
+	}  
 
 	public double getMissPercent() {
 		return response.getMissPercent();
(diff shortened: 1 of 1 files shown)


mjmjelde:
isFailed is just a helper method to help a user easily check if the CatchStatus was failed or not.  He is still able to grab the status of the catch no matter what.  I am confused as to why this function needs changing

Hindi:
let's say I want to check if I catched my pokemon.

I'd write something like
`if(catchResult != null && !catchResult.isFailed())
{
    //dostuff
}`
`else { //check why it failed }`

But when it failed you'd like to know why, maybe it is response == null, maybe it is a relevent information you want to have like CatchStatus.CATCH_FLEE. So you call `catchResult.getStatus();`
But if you do that and response == null, you have a null exception because this function uses `response.getStatus()`

mjmjelde:
So I see what you are saying now, but I still think that isFailed needs to stay the same.  It makes no sense to just have it return if `response == null`  since it could still be failed with more than just that.  Honestly just add a `@Getter` to the response and then you can do a `catchResult.getResponse() != null`

On a side note, the only time response should be null is when the encounter with the pokemon failed.

Hindi:
Ok, if using a check on the encounter is enough, I don't think that adding a @Getter is necessary. I'll try that way.

PR #156, opened 2016-07-23
Replace gson with moshi to improve performance and reduce library size

FabianTerhorst:
Moshi is depending on okio as well as okhttp. More informations: https://github.com/square/moshi


FabianTerhorst:
sorry, wrong branch.

Grover-c13:
this was the right branch, we are switching to a stable master, and a development branch :P

FabianTerhorst:
ah ok, because travis doesn´t started building.

Grover-c13:
True will need to fix that.

FabianTerhorst:
I have to add a commit i think to trigger travis again?

Grover-c13:
yeah to inc the latest config

PR #159, opened 2016-07-23
Gdev

Grover-c13:
Added using items on CapturablePokemon, added overloads for razberrys


Grover-c13:
Tested and working, razz berry gives a 1.5 multiplier to catch chance too for those wondering :D

PR #165, opened 2016-07-23
Refresh the OAuth token when it expires

rajulbhatnagar:
**Changes made:**
- Refresh OAuth token when token expires.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
-#Sun Jul 17 19:51:46 PDT 2016
+#Sat Jul 23 04:51:41 PDT 2016
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.13-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.13-all.zip
--- a/src/main/java/com/pokegoapi/auth/GoogleLogin.java
+++ b/src/main/java/com/pokegoapi/auth/GoogleLogin.java
@@ -29,16 +29,47 @@ import java.io.IOException;
 import java.net.URISyntaxException;
 
 public class GoogleLogin extends Login {
-	public static final String SECRET = "NCjF1TLi2CcY6t5mt0ZveuL7";
-	public static final String CLIENT_ID = "848232511240-73ri3t7plvk96pj4f85uj8otdat2alem.apps.googleusercontent.com";
-	public static final String OAUTH_ENDPOINT = "https://accounts.google.com/o/oauth2/device/code";
-	public static final String OAUTH_TOKEN_ENDPOINT = "https://www.googleapis.com/oauth2/v4/token";
+
 	private static final String TAG = GoogleLogin.class.getSimpleName();
 
 	private final OkHttpClient client;
 
 	public GoogleLogin(OkHttpClient client) {
 		this.client = client;
+	}
+
+	/**
+	 * Given the refresh token fetches a new access token and returns AuthInfo.
+	 * @param refreshToken Refresh token returned during initial login
+	 * @return Refreshed AuthInfo object
+	 * @throws IOException If the network call fails
+	 */
+	public AuthInfo refreshToken(String refreshToken) throws IOException {
+		HttpUrl url = HttpUrl.parse(GoogleLoginSecrets.OAUTH_TOKEN_ENDPOINT).newBuilder()
+				.addQueryParameter("client_id", GoogleLoginSecrets.CLIENT_ID)
+				.addQueryParameter("client_secret", GoogleLoginSecrets.SECRET)
+				.addQueryParameter("refreshToken", refreshToken)
+				.addQueryParameter("grant_type", "refreshToken")
+				.build();
+		//Empty request body
+		RequestBody reqBody = RequestBody.create(null, new byte[0]);
+		Request request = new Request.Builder()
+				.url(url)
+				.method("POST", reqBody)
+				.build();
+
+		Response response = client.newCall(request).execute();
+
+		Moshi moshi = new Moshi.Builder().build();
+		GoogleAuthTokenJson token = moshi.adapter(GoogleAuthTokenJson.class).fromJson(response.body().string());
+		if (token.getError() == null) {
+			return null;
+		} else {
+			AuthInfo.Builder builder = AuthInfo.newBuilder();
+			builder.setProvider("google");
+			builder.setToken(AuthInfo.JWT.newBuilder().setContents(token.getIdToken()).setUnknown2(59).build());
+			return builder.build();
+		}
 	}
 
 	/**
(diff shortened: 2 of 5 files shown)


rajulbhatnagar:
Hold off merging. I see some issues after the rebase to use moshi.

rajulbhatnagar:
Works like a dream now.

FabianTerhorst:
i think you can update to gradle 2.14.1 as well

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Sun Jul 17 19:51:46 PDT 2016
+#Sat Jul 23 04:51:41 PDT 2016
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.13-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.13-all.zip
```
langerhans, line 6: Why? That downloads sources and docs for Gradle that no one in our case really needs.

FabianTerhorst:
whats about a okhttp request interceptor to update the token when expired?

src/main/java/com/pokegoapi/auth/PtcLogin.java
```diff
@@ -166,7 +166,7 @@ public AuthInfo login(String username, String password) throws LoginFailedExcept
 					.addQueryParameter("client_id", CLIENT_ID)
 					.addQueryParameter("redirect_uri", REDIRECT_URI)
 					.addQueryParameter("client_secret", CLIENT_SECRET)
-					.addQueryParameter("grant_type", "refresh_token")
+					.addQueryParameter("grant_type", "refreshToken")
```
langerhans, line 169: Are you sure this change is required?

src/main/java/com/pokegoapi/auth/GoogleLogin.java
```diff
@@ -16,9 +16,11 @@
 package com.pokegoapi.auth;
 
 import POGOProtos.Networking.Envelopes.RequestEnvelopeOuterClass.RequestEnvelope.AuthInfo;
+
```
FabianTerhorst, line 19: are these spaces still valid to the codestyle?

gradle/wrapper/gradle-wrapper.properties
rajulbhatnagar, line 6: IDE code completion for gradle does not work without that.

src/main/java/com/pokegoapi/auth/PtcLogin.java
FabianTerhorst, line 169: the official client is using refresh_token so will should keep it .addQueryParameter("grant_type", "refresh_token")

src/main/java/com/pokegoapi/auth/GoogleLoginSecrets.java
```diff
@@ -0,0 +1,14 @@
...
+ * Created by RajulB on 7/23/2016.
+ */
+
+public class GoogleLoginSecrets {
+	public static final String SECRET = "NCjF1TLi2CcY6t5mt0ZveuL7";
+	public static final String CLIENT_ID = "848232511240-73ri3t7plvk96pj4f85uj8otdat2alem.apps.googleusercontent.com";
+	public static final String OAUTH_ENDPOINT = "https://accounts.google.com/o/oauth2/device/code";
+	public static final String OAUTH_TOKEN_ENDPOINT = "https://www.googleapis.com/oauth2/v4/token";
+
+	public static String refresh_token = null;
```
langerhans, line 13: You should add documentation to this. As soon as the application is closed this memory is cleared and the token is gone. Doing it this way requires the application to set this static field before logging back in, assuming they also acquired the initial access token.

rajulbhatnagar:
@FabianTerhorst Dont see how we can do that because we never receive a Token Expired 401. We only receive a response from niantic in the RPC which fails.

gradle/wrapper/gradle-wrapper.properties
langerhans, line 6: Well yeah, but do we need that?

src/main/java/com/pokegoapi/auth/GoogleLogin.java
rajulbhatnagar, line 19: It does pass

gradle/wrapper/gradle-wrapper.properties
rajulbhatnagar, line 6: I dont see why not. This is not bundled anyway. And its a one time download of a few megs more.

FabianTerhorst:
@rajulbhatnagar thats true right now this isn´t possible with the current implementation. When you revert the gradle wrapper change. this could be ready to merge. Im only unsure about the spacings at the imports.

src/main/java/com/pokegoapi/auth/GoogleLoginSecrets.java
rajulbhatnagar, line 13: I have added a new login method which takes care of this. Ideally the old login method with just token_id as argument should be deprecated. Even if the user does not have a fresh token_id the new method would refresh it for him.

jabbink, line 13: Please add the GPL notice to the top of this (all new) file.

src/main/java/com/pokegoapi/main/RequestHandler.java
```diff
@@ -129,8 +133,21 @@ public void sendServerRequests(ServerRequest... serverRequests) throws RemoteSer
 				lastAuth = responseEnvelop.getAuthTicket();
 			}
 
-			if (responseEnvelop.getStatusCode() == 102) {
-				throw new LoginFailedException();
+			if (responseEnvelop.getStatusCode() == 102 && GoogleLoginSecrets.refresh_token != null) {
```
jabbink, line 136: Maybe refactor the Login interface and implement this for Ptc as well, right now, by some bug in their checks of the token, those tokens are valid for a very long time, but I don't expect it to stay that way.

Then we can deprecate logging in with the token for Ptc as well and use some generic token refreshing code here (if `refreshToken` would be a part of the interface, the implementation for `PtcLogin` could currently be to return the old token).

src/main/java/com/pokegoapi/auth/GoogleLogin.java
rajulbhatnagar, line 19: Those new lines are there to segregate different import categories. XML_LEGACY_SETTINGS_IMPORTED check-style rule allows it and I believe the default checkstyle rule requires it.

rajulbhatnagar:
@FabianTerhorst The reason I created a new file was to prevent people using the API to shoot themselves on the foot. If the static was inside the GoogleLogin class and someone holds a reference to it, it would cause the login object to leak and in mobile environment it could easily mean leaking a fragment or activity!

Because I was separating this out it also made sense to move the constants to the file. We should definitely leave the static refreshToken in a separate file. Other constants can be moved back but I dont see the upside.

Ideal handling of this will require quite a bit of code refactoring where the auth provider is passed to the api client instead of just the auth tokenid and the auth provider is then self managing but I think this works just as well.

src/main/java/com/pokegoapi/main/RequestHandler.java
rajulbhatnagar, line 136: Ya the login code/logic could do with some refactoring but I dont think i'd be able to do it now. I really wanted to implement this feature as it makes google auth kinda useless without it.

My ideal API design would be to pass
an implementation of Login to PokemonGo and let the implementation handle tokens transparently.
RequestManager should just ask the PokemonGo object for AuthInfo when it wants to do a call.

FabianTerhorst:
Have tested it and looks good to me. I think the refactoring should happen after the merge. :thumbsup:

PR #166, opened 2016-07-23
add query for hatched eggs / rewrite method around EggPokemon

vmarchaud:


stirante:
Are you sure you get correct data which is returned by queryHatchedEggs? I suspect the proto file may be a bit wrong since when I tried same code and then i returned in loop all getPokemonId(int) it returned pretty random numbers. Other than that hatching egg is correct since I tried it.

src/main/java/com/pokegoapi/api/inventory/EggIncubator.java
```diff
@@ -64,9 +64,7 @@ public int getUsesRemaining() {
 	 */
 	public UseItemEggIncubatorResponse.Result hatchEgg(EggPokemon egg)
 			throws LoginFailedException, RemoteServerException {
-		if (!egg.getIsEgg()) {
-			return null;
-		}
+
```
FabianTerhorst, line 67: to match the style of the project you should remove this space

src/main/java/com/pokegoapi/api/pokemon/HatchedEgg.java
```diff
@@ -0,0 +1,14 @@
+package com.pokegoapi.api.pokemon;
```
FabianTerhorst, line 1: add the license header to the new files

FabianTerhorst, line 1: for example here: https://github.com/Grover-c13/PokeGOAPI-Java/blob/Development/src/main/java/com/pokegoapi/auth/GoogleAuthJson.java

jabbink:
the PokemonId(Long) is indeed random

maybe this helps https://github.com/Grover-c13/PokeGOAPI-Java/pull/172/files#diff-7f5d4aacf5d1aa75a91dfb222f88ef14R93

stirante:
Also sometimes using this i get "While parsing a protocol message, the input ended unexpectedly in the middle of a field.  This could mean either that the input has been truncated or that an embedded message misreported its own length."

The proto file is definetly wrong.

Grover-c13:
The issue might be from the proto using uint64 instead of fixed64, pokemon ids always seem to be fixed 64 and having them as uint seems to cause issues.

vmarchaud:
`While parsing a protocol message, the input ended unexpectedly in the middle of a field. This could mean either that the input has been truncated or that an embedded message misreported its own length."`
These packets are running error free with my bot.
Will look on for proto file.

PR #172, opened 2016-07-24
get hatch results

jabbink:
**Changes made:**
- allow retrieving hatch results
- get pokemon by ID (eg. from Hatch Result)


jabbink:
Ignore, somehow missed #166

PR #174, opened 2016-07-24
Runtimeexceptions

pvanassen:
I've changed the default Exception to a specific PokemonGoApiException which itself extends RuntimeException instead of Exception. 

This allows for easier exception checking and makes the API usable in the Java8 stream API. 

Change is backwards compatible with existing code and API implementations.

### DIFF
--- a/src/main/java/com/pokegoapi/exceptions/InvalidCurrencyException.java
+++ b/src/main/java/com/pokegoapi/exceptions/InvalidCurrencyException.java
@@ -15,7 +15,7 @@
 
 package com.pokegoapi.exceptions;
 
-public class InvalidCurrencyException extends Exception {
+public class InvalidCurrencyException extends PokemonGoApiException {
 	public InvalidCurrencyException() {
 		super();
 	}
--- a/src/main/java/com/pokegoapi/exceptions/LoginFailedException.java
+++ b/src/main/java/com/pokegoapi/exceptions/LoginFailedException.java
@@ -15,7 +15,7 @@
 
 package com.pokegoapi.exceptions;
 
-public class LoginFailedException extends Exception {
+public class LoginFailedException extends PokemonGoApiException {
 	public LoginFailedException() {
 		super();
 	}
--- a/src/main/java/com/pokegoapi/exceptions/NoSuchItemException.java
+++ b/src/main/java/com/pokegoapi/exceptions/NoSuchItemException.java
@@ -15,7 +15,7 @@
 
 package com.pokegoapi.exceptions;
 
-public class NoSuchItemException extends Exception {
+public class NoSuchItemException extends PokemonGoApiException {
 	public NoSuchItemException() {
 		super();
 	}
--- a/src/main/java/com/pokegoapi/exceptions/PokemonGoApiException.java
+++ b/src/main/java/com/pokegoapi/exceptions/PokemonGoApiException.java
@@ -1,0 +1,14 @@
+package com.pokegoapi.exceptions;
+
+public class PokemonGoApiException extends RuntimeException {
+	public PokemonGoApiException() {
+	}
+
+	public PokemonGoApiException(String message) {
+		super(message);
+	}
+
+	public PokemonGoApiException(Throwable cause) {
+		super(cause);
+	}
+}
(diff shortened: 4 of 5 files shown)


jabbink:
When I introduced that exception I made it an Exception on purpose, because RuntimeExceptions are bad because no IDE suggests you check those exceptions at all and then at ... runtime, stuff suddenly starts breaking.

IMO RuntimeExceptions are the root of all evil (looking at you Sun/Oracle with your `NumberFormatException`, same type of exception as a FileNotFound/IOException, but that one is an `Exception`)

pvanassen:
You may be right but Sun/Oracle (and Microsoft's .NET) decided unchecked is the way to go. I think using checked exceptions is nice because of IDE support but it limits the usage of the API.

Possibly we can devide the exceptions into two categories, first, user has to act, like login errors. The second, user cannot help it, like server errors as suggested here: 
https://docs.oracle.com/javase/tutorial/essential/exceptions/runtime.html

jabbink:
My point exactly, they're highly inconsistent (eg IOException and FileNotFoundException)

> user has to act, like login errors

needs many codechanges then as right now we don't know why a Login error occured

rajulbhatnagar:
I personally prefer exception unless you reach a unrecoverable state where it would make sense to crash(Should happen only during testing).

And IDE checks are another big thing as most people don't read documentation. We'll have a lot of random crashes with RuntimeException. Just MHO.

jabbink:
You don't have random crashes when it's `Exception` as you HAVE to catch it.

pvanassen:
Totally disagree. With checked exceptions people just wrap everything into a big try catch and let it crash. Same as with RuntimeException, just with more code. If you're worried about random crashes you must make sure your logging is up to date. This will help people understand why things happen, leaving it up to them wether to try-catch in that particular situation. 

Making everything a checked exception because people don't read documentation and it will introduce random crashes is a sophism.

LunNova:
The problem is that most of the time, when a checked exception occurs, there's nothing useful you can do, you just want to give up, so you wrap it in a RuntimeException and rethrow it, achieving nothing. It's an annoying extra bit of boilerplate code.

Checked exceptions are particularly annoying when working with java 8 streams.

@pvanassen If you want to be able to treat a checked exception as a runtime exception, you could use lombok's `@SneakyThrows` annotation: https://projectlombok.org/features/SneakyThrows.html

edit: removed silly snarky comment

davinkevin:
+1 and oracle also create unchecked version of old checkedException ( ex Unchecked IOException). 

IMO, it's simple to work without it... A lot of lib follow this pattern, and if you want to avoid it, you can use Try.of Javaslang to catch it automaticlly...

Aphoh:
Just saying, a lot of the time Java 8 streams are not the ideal format when it comes to exception handling. Things like RxJava tend to handle this a bit more elegantly, but it is still not perfect. That being said, it is not our responsibility for people wrapping everything in a big try/catch block. It is our responsibility to make the APIs easy to use correctly and difficult to use incorrectly. Having runtime exceptions makes it very very easy to misuse.

pvanassen:
Ok, I'm going to leave the checked exceptions. I'll just keep my own repo up to date and use that.

LunNova:
[Does Java need Checked Exceptions?](https://web.archive.org/web/20160102073954/http://www.mindview.net/Etc/Discussions/CheckedExceptions)

[Java's checked exceptions were a mistake (and here's what I would like to do about it)](http://radio-weblogs.com/0122027/stories/2003/04/01/JavasCheckedExceptionsWereAMistake.html)

[The Trouble with Checked Exceptions | A Conversation with Anders Hejlsberg](http://www.artima.com/intv/handcuffs.html)

jabbink:
Oh nice, cherry picked stuff that confirms your opinion. I could do that.

EDIT: eg http://stackoverflow.com/a/614494/1834301

Also I fully agree with @Aphoh 

> It is our responsibility to make the APIs easy to use correctly and difficult to use incorrectly. Having runtime exceptions makes it very very easy to misuse.

LunNova:
> cherry-picked

I'm not going to send links which say checked exceptions are good if I'm trying to convince you that they're not! Better to provide counter-arguments than to complain about cherry-picking here.

I chose those links as they're referenced when discussing why Kotlin [doesn't have checked exceptions](https://kotlinlang.org/docs/reference/exceptions.html).

PR #175, opened 2016-07-24
Added maven artifact support

pvanassen:
Added maven publish support to allow us legacy maven users to easily allow using the api.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,6 +17,7 @@ apply plugin: 'java'
 apply plugin: 'maven'
 apply plugin: 'com.google.protobuf'
 apply plugin: 'checkstyle'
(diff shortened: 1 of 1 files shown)


langerhans:
Is this for the sources jar and if so will it now be built twice? Cause I added the sources export just yesterday above your change.

pvanassen:
No, this is to publish the compiled and sources jar together with the pom.xml into your local .m2 or a remote maven repository. It should not do things twice

langerhans:
This already happens when you run `./gradlew install`. Don't think that change is required for this.

pvanassen:
Alright, in that case you can close it. Sorry for the inconvenience.

PR #198, opened 2016-07-25
Implemented slf4j

pvanassen:
Implemented SLF4j. 

At first I attempted to reuse the existing Log class but this added extra information in the stack traces which was unnecessary. With SLF4j people implementing the API can choose which logging framework they want to use and are not stuck to System.out output. Lombok also integrates nicely with it through the @Slf4j annotation.


FabianTerhorst:
Im not seeing any advantages here to add 600 more methods to this library just by adding a logging library.

pvanassen:
Moved it to https://github.com/Grover-c13/PokeGOAPI-Java/pull/199. Forgot to first merge, too many merge conflicts

PR #199, opened 2016-07-25
Implemented slf4j

pvanassen:
Implemented SLF4j.

At first I attempted to reuse the existing Log class but this added extra information in the stack traces which was unnecessary. With SLF4j people implementing the API can choose which logging framework they want to use and are not stuck to System.out output. Lombok also integrates nicely with it through the @Slf4j annotation.

Attempt 2, fixed merge conflicts (old pr: https://github.com/Grover-c13/PokeGOAPI-Java/pull/198)


FabianTerhorst:
from #198: 
Im not seeing any advantages here to add 600 more methods to this library just by adding a logging library.
 Something like a static method to add a different logger would could be a solutions for that.

FabianTerhorst:
In android you use for example https://github.com/JakeWharton/timber or the default https://developer.android.com/reference/android/util/Log.html

pvanassen:
Very good very well but why did you write your own logger then instead of using one of the two examples?
You are writing an API, not some Java code people include directly right?

langerhans:
Cause this library isn't exclusively used on Android.

I'm indifferent about the PR. With the protos alone taking up 28k methods, it doesn't seem to add much with 600.

FabianTerhorst:
With a custom logger you could wrap the log calls to the platform specific apis.

pvanassen:
That is exactly what SFL4j is, a logger facade to log to specific API's, even to Android logging: 
http://www.slf4j.org/android/

FabianTerhorst:
But this could implemented with just an logger interface and a static method to to set the logger interface at the start of the program.

pvanassen:
Ok, I'm going to leave you with your own logger. I'll just keep my own repo up to date and use that.

Aphoh:
@pvanassen There is no benefit to adding sl4j, and it just adds bulk to an already big library. Implement the Logger class and use sl4j if you like in your projects, but it doesn't belong in this lib.

PR #203, opened 2016-07-25
WIP: Gym/Battle

Grover-c13:
WIP: Battle + Gym API

### DIFF
--- a/src/main/java/com/pokegoapi/api/gym/Battle.java
+++ b/src/main/java/com/pokegoapi/api/gym/Battle.java
@@ -1,0 +1,186 @@
+/*
+ *     This program is free software: you can redistribute it and/or modify
+ *     it under the terms of the GNU General Public License as published by
+ *     the Free Software Foundation, either version 3 of the License, or
+ *     (at your option) any later version.
+ *
+ *     This program is distributed in the hope that it will be useful,
+ *     but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *     GNU General Public License for more details.
+ *
+ *     You should have received a copy of the GNU General Public License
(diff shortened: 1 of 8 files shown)


Grover-c13:
Just incase someone else decides to start work on it while im not on my PC :P.

Currently some issues with it, i am guessing to do it properly you need to start a battle, then join a battle. Currently getting a bad result returned though so really need some network dumps of a full gym battle.

src/main/java/com/pokegoapi/api/gym/Battle.java
```diff
@@ -0,0 +1,186 @@
...
+		for (int i = 0; i < team.length; i++) {
+			builder.addAttackingPokemonIds(team[i].getId());
+		}
+
+
+		List<PokemonDataOuterClass.PokemonData> defenders = gym.getDefendingPokemon();
+		builder.setGymId(gym.getId());
+		builder.setPlayerLongitude(api.getLongitude());
+		builder.setPlayerLatitude(api.getLatitude());
+		builder.setDefendingPokemonId(defenders.get(0).getId()); // may need to be sorted
```
vmarchaud, line 92: L#78 : they are already from what i saw, but you may check if the gym is actually defended (no team has claim it), its could lead to a NPE

src/main/java/com/pokegoapi/api/gym/Battle.java
```diff
@@ -0,0 +1,186 @@
...
+		}
+
+	}
+
+	public Result start() throws LoginFailedException, RemoteServerException {
+
+		Builder builder = StartGymBattleMessageOuterClass.StartGymBattleMessage.newBuilder();
+
+		for (int i = 0; i < team.length; i++) {
+			builder.addAttackingPokemonIds(team[i].getId());
```
vmarchaud, line 84: L#70 : You could just use addAllAttackingPokemonIds() methods

vmarchaud:
Network dump of a gym attack: http://pastebin.com/enDzFvUN

PR #206, opened 2016-07-25
[DO NOT MERGE] 	split sample and library into different modules

FabianTerhorst:
- split sample and library into different modules
- update okhttp and lombok

Todo:
readd protobuf commit
fix build errors caused from incompatible proto files


src/resources/protobuf
```diff
@@ -1 +0,0 @@
-Subproject commit ff99b436a34f0db79e3f1db805811f083405aecd
```
FabianTerhorst: i don´t know why this is happening

library/src/resources/protobuf
```diff
@@ -0,0 +1 @@
+Subproject commit ff99b436a34f0db79e3f1db805811f083405aecd
```
FabianTerhorst: The submodule isn´t correctly added

Grover-c13:
I added another example btw :p

FabianTerhorst:
@Grover-c13 Then problem is that i can´t figure out how to reinit the submodule after changing the path.

FabianTerhorst:
Maybe the easiest way to merge this would be recreate the branch.

Grover-c13:
yeah maybe :P

PR #207, opened 2016-07-25
Added iv calculator

Jari27:
**Changes made:** Added an IV Ratio calculator to the Pokemon class. Use like this

`double ivRatio = pokemon.getIVRatio()`

**Reason for change**: seemed useful.

### DIFF
--- a/src/main/java/com/pokegoapi/api/pokemon/Pokemon.java
+++ b/src/main/java/com/pokegoapi/api/pokemon/Pokemon.java
@@ -270,7 +270,7 @@ public class Pokemon {
 	
 	/**
 	 * Calculates the pokemons IV ratio.
-	 * @return the pokemons IV ratio as a double between 0 and 1.0, 1.0 being perfect IVs.
+	 * @return the pokemons IV ratio as a double between 0 and 1.0, 1.0 being perfect IVs
 	 */
 	public double getIVRatio() {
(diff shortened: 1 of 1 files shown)


FabianTerhorst:
Could be helpful: http://www.pokewiki.de/Determinant_Values

src/main/java/com/pokegoapi/api/pokemon/Pokemon.java
```diff
@@ -267,6 +267,14 @@ public int getIndividualDefense() {
 	public int getIndividualStamina() {
 		return proto.getIndividualStamina();
 	}
+	
+	/**
+	 * Calculates the pokemons IV ratio
```
jabbink: Doesn't this break `checkstyleMain` because it does not contain a dot?

FabianTerhorst:
http://bulbapedia.bulbagarden.net/wiki/Individual_values

src/main/java/com/pokegoapi/api/pokemon/Pokemon.java
Jari27: I can't get checkstyle to run. If I need to change it I will. I can also start working on annotating more of the classes.

FabianTerhorst: just change `Calculates the pokemons IV ratio` to `Calculates the pokemons IV ratio.`

FabianTerhorst:
only at the title, not at the return description.

FabianTerhorst:
👍

PR #211, opened 2016-07-25
use lombok for PtcAuthJson and PtcError

FabianTerhorst:


src/main/java/com/pokegoapi/auth/PtcError.java
```diff
@@ -15,8 +15,13 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
 public class PtcError {
 
+	@Getter
+	@Setter
 	private String error;
 
 	public String getError() {
```
langerhans: I guess those can be removed then?

FabianTerhorst: did already forgot to sync^^

FabianTerhorst:
im opening another pr to resolve conflicts

PR #217, opened 2016-07-25
Added request to add modifiers to forts, for example a lure-module on a pokestop

aschlosser:
**Changes made:**

Added a method `addModifier(ItemId)` to the `Pokestop` class which instantiates and sends a `AddFortModifierMessage` to the server. It is used to add the modifier specified as an ItemId to the pokestop, usually this modifier will be a lure module (`ITEM_TROY_DISK`).

### DIFF
--- a/src/main/java/com/pokegoapi/api/map/fort/Pokestop.java
+++ b/src/main/java/com/pokegoapi/api/map/fort/Pokestop.java
@@ -15,10 +15,13 @@
 
 package com.pokegoapi.api.map.fort;
 
+import POGOProtos.Inventory.Item.ItemIdOuterClass;
 import POGOProtos.Map.Fort.FortDataOuterClass;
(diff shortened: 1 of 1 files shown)


src/main/java/com/pokegoapi/api/map/fort/Pokestop.java
```diff
@@ -127,6 +130,29 @@ public PokestopLootResult loot() throws LoginFailedException, RemoteServerExcept
 	}
 
 	/**
+	 * Adds a modifier to this pokestop. (i.e. add a lure module)
+	 *
+	 * @param item the modifier to add to this pokestop
+	 */
```
Jari27, line 138: Can you add `@throws` ?

aschlosser, line 138: Whoops. Fixed it.

PR #220, opened 2016-07-25
Add cause to LoginFailedExceptions.

TwoStone:
**Changes made:**
- This pr adds the cause of a failed login to the LoginFailedException as the cause. This is done for make debugging failed logins easier.
- Implemented for both login types: PTC and Google

Signed-off-by: Niklas Walter walter.niklas@gmail.com


FabianTerhorst:
Nice, great idea.

PR #224, opened 2016-07-25
getItemsCount function

ramarro123:
**Changes made:**
added a method to get the used spaces on bag.
useful to trigger actions based on remaining space on the bag


Jari27:
Can you document it? Otherwise someone else will have to do it some day.

Also, I think checkStyleMain doesn't allow one character variables. Maybe change it to `Item item`?

ramarro123:
here we go

williamho:
isn't this the same as `items.size()`?

Deagler:
No, It isn't the same. That isn't the way the ItemBag works.

williamho:
oh sorry, I misread

PR #226, opened 2016-07-25
tweak checkstyle config to be friendly

vmarchaud:
- Allow condition with only one line to remove the bracket
- Forbid empty method/loop/constructors (they shouldnt exist if they are empty)
- Remove the rule that forbid this : getIVRatio (the V is not allowed in capslock)
- Remove the stupid rule that force an order in import
- Remove the rule that force to write a dot at the end of the firstline of javadoc.

### DIFF
--- a/config/checkstyle.xml
+++ b/config/checkstyle.xml
@@ -57,7 +57,6 @@
             <property name="option" value="TEXT"/>
             <property name="tokens" value="LITERAL_TRY, LITERAL_FINALLY, LITERAL_IF, LITERAL_ELSE, LITERAL_SWITCH"/>
         </module>
-        <module name="NeedBraces"/>
         <module name="LeftCurly">
             <property name="maxLineLength" value="120"/>
         </module>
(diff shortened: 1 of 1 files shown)


ramarro123:
please merge it immediately :D or remove checkstyle, i am starting to hate it 💃 
hehehe

@Sadye  lol, same time, totally different opinions

Jari27:
I actually like the first rule you tweaked, makes for prettier and less messy code, especially if you update it later and forget it's not in a block. (On the other hand, how often has this really happened..?)

Opinions?

vmarchaud:
I dont use bracket for one line since, its more relevant as i personal preference than an actual important style of coding, so i think we must allow everyone to use bracket as they want if its compiling. We should be open to everyone prefered coding style that isnt relevant at the readability of the code itself.
I actually found myself putting more time into style checking than coding, that doesnt sound like interesting for new contributor.

Waiting @Grover-c13 @jabbink to share their view on this.

Jari27:
In the end it's also important that all code is written the same way, so it forms a cohesive api.

Imagine if you named your functions getIVRatio(), while someone else named them calculateIVRatio(), and that for all functions.

It'd be very hard for end users to pick the api up quickly.

vmarchaud:
@Sadye i dont get why you are talking about the function name since we are reviewing all PR, plus a coding checkstyle will not forbid that.

Jari27:
I know. It was the first example I could think of from the top of my head. Bad example at that, I admit.

I was trying to point out why having the checkStyle module is usually a good thing.

vmarchaud:
I'm agree that is important to have the checkstyle, but i dont think its a good idea to make people put more effort into useless styling (ie : force brace of one line of condition block) than coding.

kyleshrader:
I personally hate one line ifs (no braces)... It's not very forward thinking and adds potential for bugs in updates - but lets agree this is a holy war that we should not get into... do what you will w/ checkstyles :P

PR #234, opened 2016-07-26
Rewrote and Refactored login logic.Credential Providers transparently handle token refreshing for both Ptc and Google providers

rajulbhatnagar:
**Changes made:**
- Transparent token handling by the credential provider. Request handler does not need to keep track of authinfo
- Token refresh for Google and Ptc logins
- All the API's should work. A single line change to login code required.
- Proper exception handling


FabianTerhorst:
nice refactoring.

Grover-c13:
Tested and checkStyled?

jabbink:
Will test it now

Issue #237, opened 2016-07-26
Correct inUse for incubators


jabbink:
**Description:**
The property `inUse` is not set correctly in the API, it returns `false` at all times right now.

**Expected behavior:**
Return `true` when an incubator is occupied.

**Actual behavior:**
Always not in use

**Proposed fix:**
Remove the `inUse` field and add this function

``` java
    public boolean isInUse() {
        return getKmTarget() > pgo.getPlayerProfile().getStats().getKmWalked();
    }
```

Maybe add some error handling to make sure the playerProfile and stats are actually set.

Jari27:
Isn't this the wrong way around? I think the incubator is in use if `getKmTarget() > pgo.getPlayerProfile().getStats().getKmWalked();`

jabbink:
yes, indeed the wrong way around. Copypasted the sample code from one of my projects, that filters unused incubators. Fixing the top comment.

PR #238, opened 2016-07-26
Iss237: fix isInUse() always returning false

Jari27:
**Fixed issue:** #237

**Changes made:** 
Removed the field inUse,
added a new method that actually checks whether it is being used right now.

(same changes as proposed in #237)


vmarchaud:
Will make a PR with same fix + egg walked distance, i bother you if we close this one and we merged mine ? CF #240

PR #262, opened 2016-07-26
Hypno should actually be in FAMILY_DROWZEE

Sheigutn:
^ See title


Grover-c13:
haha i completely forgot drowzee evolves, i just thought they where both separate pokemon, thank you!

PR #265, opened 2016-07-27
Automatic google login with username and password

svarzee:
**Changes made:**
- Added gpsoauth dependency
- Created a credential provider that can be used with username and password and doesn't require manual interaction

### DIFF
--- a/src/main/java/com/pokegoapi/auth/GoogleAutoCredentialProvider.java
+++ b/src/main/java/com/pokegoapi/auth/GoogleAutoCredentialProvider.java
@@ -21,39 +21,32 @@ public class GoogleAutoCredentialProvider extends CredentialProvider {
 	private static String GOOGLE_LOGIN_APP = "com.nianticlabs.pokemongo";
 	private static String GOOGLE_LOGIN_CLIENT_SIG = "321187995bc7cdc2b5fc91b11a96e2baa8602c62";
 
+	private final Gpsoauth gpsoauth;
 	private final String username;
-	private final String password;
+	private TokenInfo tokenInfo;
 
-	private AuthToken authToken;
-	private AuthInfo authInfo;
 
 	/**
 	 * Constructs credential provider using username and password
 	 *
 	 * @param username - google username
 	 * @param password - google password
-	 * @throws LoginFailedException - login failed possibly due to invalid credentials
+	 * @throws LoginFailedException  - login failed possibly due to invalid credentials
 	 * @throws RemoteServerException - some server/network failure
 	 */
-	public GoogleAutoCredentialProvider(String username, String password)
+	public GoogleAutoCredentialProvider(OkHttpClient httpClient, String username, String password)
 			throws LoginFailedException, RemoteServerException {
+		this.gpsoauth = new Gpsoauth(httpClient);
 		this.username = username;
-		this.password = password;
-		login(username, password);
+		this.tokenInfo = login(username, password);
 	}
 
-	private void login(String username, String password) throws RemoteServerException, LoginFailedException {
+	private TokenInfo login(String username, String password) throws RemoteServerException, LoginFailedException {
 		try {
-			OkHttpClient httpClient = new OkHttpClient();
-			Gpsoauth gpsoauth = new Gpsoauth(httpClient);
-			this.authToken = gpsoauth.login(
-					username,
-					password,
-					GOOGLE_LOGIN_ANDROID_ID,
-					GOOGLE_LOGIN_SERVICE,
-					GOOGLE_LOGIN_APP,
-					GOOGLE_LOGIN_CLIENT_SIG
-			);
+			String masterToken = gpsoauth.performMasterLoginForToken(username, password, GOOGLE_LOGIN_ANDROID_ID);
+			AuthToken authToken = gpsoauth.performOAuthForToken(username, masterToken, GOOGLE_LOGIN_ANDROID_ID,
+					GOOGLE_LOGIN_SERVICE, GOOGLE_LOGIN_APP, GOOGLE_LOGIN_CLIENT_SIG);
+			return new TokenInfo(authToken, masterToken);
 		} catch (IOException e) {
 			throw new RemoteServerException(e);
 		} catch (Gpsoauth.TokenRequestFailed e) {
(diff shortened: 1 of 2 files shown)


Grover-c13:
nice

FabianTerhorst:
nice library you found there.

FabianTerhorst:
@svarzee are you sure the jar is added correctly? Travis build is failing, because of missing package.

svarzee:
@FabianTerhorst hmm, I'll figure it out tomorrow, I haven't slept for a night to port gpsaouth

svarzee:
@FabianTerhorst feel free to fix it while I am sleeping ;)

vmarchaud:
I dont see any refresh token method in your lib, you should implement it so dont store in RAM the google credentials.
Plus we are using gradle, i think you should make a jitpack like us to host it somewhere.

rajulbhatnagar:
I guess this makes it convenient so its a good option to have in the API. But I personally dont like it too much as I am not comfortable entering google login credentials for people using primary accounts.

And it does not work with 2 factor authentication. Also I dont see any library implement token refresh :/  rather than re-logging in generating a new master token

Grover-c13:
i think it really depends on your use case: if your providing a service that uses its own accounts (like a pokemon scanning map) automated login would be best for them for load balancing, but anything that uses other peoples accounts should be using oauth imo.

Grover-c13:
and yes it really does need refreshing.

rajulbhatnagar:
Well its still oauth. Just someone was able to extract the master token from an old version google play services which allows us to act like the googleapiclient

rajulbhatnagar:
Hmm looking at the library code it looks like the mastertoken should be usable as a refresh token.

Grover-c13:
yeah, by that i mean interfacing with users should use proper oauth flows (ie the username and password never enter the java api)

Gospodzin:
@rajulbhatnagar yes it can be used to refresh the auth token however I need to check for how long it remains valid

svarzee:
I used the master token to refresh auth token.

AbandonedCart:
While I see the drawbacks of having the user enter their credential information, this would be extremely beneficial to Android. On Android, there are means to have the user approve the app to access accounts, meaning it could enter the username and have them enter a password without the actual app caching them locally. This would only be desired as a secondary method, though.

The most significant issue I have had with API usage has been effectively displaying the required views to have the user perform their manual steps in the process of an AsyncTask and then actually having them understand exactly how and why they had to log in with a website and enter a code.

Unfortunately, building failed when this change was included until a line length issue was resolved, but then caused building of the target application to fail. The error for that failure is included below:

`[dx] UNEXPECTED TOP-LEVEL EXCEPTION:
       [dx] java.lang.RuntimeException: Exception parsing classes
       [dx]     at com.android.dx.command.dexer.Main.processClass(Main.java:752)
       [dx]     at com.android.dx.command.dexer.Main.processFileBytes(Main.java:718)
       [dx]     at com.android.dx.command.dexer.Main.access$1200(Main.java:85)
       [dx]     at com.android.dx.command.dexer.Main$FileBytesConsumer.processFileBytes(Main.java:1645)
       [dx]     at com.android.dx.cf.direct.ClassPathOpener.processArchive(ClassPathOpener.java:284)
       [dx]     at com.android.dx.cf.direct.ClassPathOpener.processOne(ClassPathOpener.java:166)
       [dx]     at com.android.dx.cf.direct.ClassPathOpener.process(ClassPathOpener.java:144)
       [dx]     at com.android.dx.command.dexer.Main.processOne(Main.java:672)
       [dx]     at com.android.dx.command.dexer.Main.processAllFiles(Main.java:574)
       [dx]     at com.android.dx.command.dexer.Main.runMonoDex(Main.java:311)
       [dx]     at com.android.dx.command.dexer.Main.run(Main.java:277)
       [dx]     at com.android.dx.command.dexer.Main.main(Main.java:245)
       [dx]     at com.android.dx.command.Main.main(Main.java:106)
       [dx] Caused by: com.android.dx.cf.iface.ParseException: bad class file magic (cafebabe) or version (0034.0000)
       [dx]     at com.android.dx.cf.direct.DirectClassFile.parse0(DirectClassFile.java:472)
       [dx]     at com.android.dx.cf.direct.DirectClassFile.parse(DirectClassFile.java:406)
       [dx]     at com.android.dx.cf.direct.DirectClassFile.parseToInterfacesIfNecessary(DirectClassFile.java:388)
       [dx]     at com.android.dx.cf.direct.DirectClassFile.getMagic(DirectClassFile.java:251)
       [dx]     at com.android.dx.command.dexer.Main.parseClass(Main.java:764)
       [dx]     at com.android.dx.command.dexer.Main.access$1500(Main.java:85)
       [dx]     at com.android.dx.command.dexer.Main$ClassParserTask.call(Main.java:1684)
       [dx]     at com.android.dx.command.dexer.Main.processClass(Main.java:749)
       [dx]     ... 12 more
       [dx] 1 error; aborting`

src/main/java/com/pokegoapi/auth/GoogleAutoCredentialProvider.java
```diff
@@ -21,63 +21,79 @@
...
+					GOOGLE_LOGIN_SERVICE, GOOGLE_LOGIN_APP, GOOGLE_LOGIN_CLIENT_SIG);
+			return new TokenInfo(authToken, masterToken);
 		} catch (IOException e) {
 			throw new RemoteServerException(e);
 		} catch (Gpsoauth.TokenRequestFailed e) {
 			throw new LoginFailedException(e);
 		}
 	}
 
+	private TokenInfo refreshToken(String username, String refreshToken) throws RemoteServerException, LoginFailedException {
```
LoungeKatt: Line is longer than 120 characters (found 123). [LineLength]

This causes a build failure, but is easily resolved by renaming to `refToken`

benjy3gg:
can you add a quick example code?

Grover-c13:
`GoogleAutoCredentialProvider` cre = new GoogleAutoCredentialProvider(http, "username", "password");
PokemonGo go = new PokemonGo(cre, http, time);``

benjy3gg:
the Time object can be
`new Time() {
                        @Override
                        public long currentTimeMillis() {
                            return System.currentTimeMillis();
                        }
                    }`
?

Grover-c13:
Use SystemTimeImpl by default

benjy3gg:
done.
Im on android and when i try to compile i get:
Unsupported class file version 52.0

Paul776:
This causes my Android build to break

Error:Execution failed for task ':app:transformClassesWithJarMergingForLiteDebug'.

> com.android.build.api.transform.TransformException: java.util.zip.ZipException: duplicate entry: okhttp3/Address.class

Even If I exclude okio I get same error, anyone using it with android has the same problem ?

AbandonedCart:
The issue is gpsoauth-java.

The build.gradle declares `sourceCompatibility = 1.8` and should be `sourceCompatibility = 1.7` to resolve this error.

Paul776:
Yep that fixe the problem, thank you

benjy3gg:
where exactly?

AbandonedCart:
@benjy3gg https://github.com/svarzee/gpsoauth-java/pull/1

Paul776:
Thanks!

Paul776:
However now I get this while performing the login
 Caused by: java.lang.NoClassDefFoundError: Failed resolution of: Ljava/util/Base64;
which is a java8 class

AbandonedCart:
http://stackoverflow.com/a/14413290/461982

But that does not support url encoding, so it isn't a complete solution

Paul776:
I tried adding apache common codec and using Base64 from there to encode in 64 for it to work with java 7
but the login fails probably same issue that in the stack post

Paul776:
I tried  by adding apache common codec and using Base64 from there to encode in 64 for it to work with java 7

but the login does not work.

AbandonedCart:
Updated the PR using a refactored Android Base64, but bear in mind

CRITICAL

Not to mention it would appear Niantic has disabled new auth requests temporarily. It is uncertain if this is related to the excessive permissions issue or a permanent change, though.

Paul776:
Thanks for this, yes it appears I'm having trouble log-in at the moment. I will give it a go later!

AbandonedCart:
Worst case is having to start doing auth as an iPhone. The premise being that they can't completely remove auth requests without completely redesigning iOS support.

Paul776:
Have you ever managed to login successfully using this pull request ? I'm always getting 403

benjy3gg:
i cant get token too. 403

Paul776:
It is a problem with the lib somewhere as I have made a python script the can get me the token without problem.
@svarzee when you wake up from your siesta could you have a look?

Thanks for the lib by the way!

AbandonedCart:
They are currently blocking the auth requests. It's not a local issue.

benjy3gg:
are you sure @LoungeKatt ?

AbandonedCart:
Yes

benjy3gg:
is this login method different from the login method use in the app?

Grover-c13:
The automated one is the closest to the app.

AbandonedCart:
It still isn't an identical request. If the basis for your question is that you can log in using the app and not using the API, then it is important to remember that you are not signing with Niantic's keys.

Blocking third-party requests, by definition, blocks requests made outside the scope of the app itself.

You can verify this by testing the original, "manual" oath process. That is currently unavailable, as well.
If there are still doubts, the same can be achieved by attempting to log in using an iPhone and Google.
  Google login using an iPhone is an official request, signed by Niantic, but simply sits on loading.

benjy3gg:
okay... you think it will be possible to login again sometime soon?

benjy3gg:
so you cant login with iphone into the official app right now over google login?

john086:
There are news ? about instant run android ?

Any idea?

Error:Execution failed for task ':app:transformClassesWithDexForDebug'.

com.android.build.api.transform.TransformException: java.lang.RuntimeException: com.android.ide.common.process.ProcessException: java.util.concurrent.ExecutionException: com.android.ide.common.process.ProcessException: Return code 1 for dex process
Error:Error converting bytecode to dex:
Cause: Dex cannot parse version 52 byte code.
This is caused by library dependencies that have been compiled using Java 8 or above.
If you are using the 'java' gradle plugin in a library submodule add 
targetCompatibility = '1.7'
sourceCompatibility = '1.7'
to that submodule's build.gradle file.

AbandonedCart:
@benjy3gg That about sums it up.

@john086 https://github.com/svarzee/gpsoauth-java/pull/1

PR #271, opened 2016-07-27
[DONOTMERGE] Added calculations based on CP

pvanassen:
Added all sorts of calculations for CP and CP multiplier, like next CP, costs of power, which level a pokemon is, what the max CP of the pokemon is.


src/resources/protobuf
```diff
@@ -1 +1 @@
-Subproject commit 8e0fac2d685fbc635886e9091039b245bb86511c
```
pvanassen: This should be undone

PR #272, opened 2016-07-27
Added CP based calculations

pvanassen:
Added CP based calculations which allow you to see how much CP a powerup would add, how much candy/stardust it will cost, what the max cp of a pokemon is. 
Source of information: http://pokemongo.gamepress.gg/pokemon-stats-advanced


src/main/java/com/pokegoapi/api/pokemon/PokemonCpUtils.java
```diff
@@ -0,0 +1,247 @@
...
+		CPMULTIPLIER_LEVEL.put(0.764486065f, 35.5f);
+		CPMULTIPLIER_LEVEL.put(0.76739717f, 36f);
+		CPMULTIPLIER_LEVEL.put(0.770297266f, 36.5f);
+		CPMULTIPLIER_LEVEL.put(0.7731865f, 37f);
+		CPMULTIPLIER_LEVEL.put(0.776064962f, 37.5f);
+		CPMULTIPLIER_LEVEL.put(0.77893275f, 38f);
+		CPMULTIPLIER_LEVEL.put(0.781790055f, 38.5f);
+		CPMULTIPLIER_LEVEL.put(0.78463697f, 39f);
+		CPMULTIPLIER_LEVEL.put(0.787473578f, 39.5f);
+		CPMULTIPLIER_LEVEL.put(0.79030001f, 40f);
```
aschlosser: Using an approximation of this function is probably a good idea as it improves performance and readability. I took the time to use an online regression tool to compute one direction:

```
public static float f(float x) {
    float y;
    if (x < 0.734f) {
        // compute polynomial approximation obtained by regression
        y = 58.35178527f * x * x - 2.838007664f * x + 0.8539209906f;
    } else {
        // compute linear approximation obtained by regression
        y = 171.0112688f * x - 95.20425243f;
    }
    // round to nearest .5 value and return
    return Math.round((y) * 2) / 2.0f;
}
```

This function returns the same result for all data points.
The tools I used are these: [Polynomial Approximation](http://www.xuru.org/rt/PR.asp) and [Linear Approximation](http://www.xuru.org/rt/LR.asp). To see how I determined the cutoff point take a look at the function plot: 
cp-multiplier-map

For the other direction it should be possible to use the same techniques to find an approximation, I can look into it tomorrow if you need help with this.

pvanassen: You can calculate all of the above values with the forumla listed at the bottom of this article: 
http://pokemongo.gamepress.gg/cp-multiplier

But for the sake of performance and readability I choose to list them all instead of having some complex math in there.

pvanassen:
I'm getting rounding errors, and therefor map lookup is not very reliable. I think your method will help this but I also need the calculations the other way around.

src/main/java/com/pokegoapi/api/pokemon/PokemonCpUtils.java
jabbink: > cp multiplier

I didn't really read into this, but the API offers access to a "cp multiplier" for pokemon

pvanassen: Yes it does, CP multiplier and additional CP multiplier. The first one is static, the second one increases with powerups. If you add both of them together you can look up the level of the pokemon. 
With the level and powerups you can calculate candy and stardust needed.

pvanassen: I'm getting rounding errors, and therefor map lookup is not very reliable. I think your method will help this but I also need the calculations the other way around and I am not very awesome at math

src/main/java/com/pokegoapi/api/pokemon/Pokemon.java
```diff
@@ -390,6 +389,45 @@ public double getBaseFleeRate() {
...
+	public int getCandyCostsForPowerup() {
+		return PokemonCpUtils.getCandyCostsForPowerup(proto.getCpMultiplier() + proto.getAdditionalCpMultiplier(),
+				proto.getNumUpgrades());
+	}
+
+	/**
+	 * @return Cost of stardust for a powerup
+	 */
+	public int getStardustCostsForPowerup() {
+		return PokemonCpUtils.getCandyCostsForPowerup(proto.getCpMultiplier() + proto.getAdditionalCpMultiplier(),
```
stirante: I think here should be PokemonCpUtils.getStartdustCostsForPowerup(...)

pvanassen: Good catch

PR #335, opened 2016-07-29
[NOT READY]

Grover-c13:
RequestHandler runs in its own threads and handles scheduling to stop throtling, starting readding methods with using an async system, currently a little messy and some parts may seem hackey, putting this up for people to review as working on it.

### DIFF
--- a/src/main/java/com/pokegoapi/main/RequestHandler.java
+++ b/src/main/java/com/pokegoapi/main/RequestHandler.java
@@ -23,6 +23,8 @@ import com.pokegoapi.api.PokemonGo;
 import com.pokegoapi.exceptions.LoginFailedException;
 import com.pokegoapi.exceptions.RemoteServerException;
 import com.pokegoapi.util.Log;
+import lombok.Getter;
+import lombok.Setter;
 import okhttp3.OkHttpClient;
 import okhttp3.RequestBody;
 import okhttp3.Response;
@@ -33,17 +35,20 @@ import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Random;
+import java.util.concurrent.ConcurrentLinkedQueue;
 
-public class RequestHandler {
+public class RequestHandler implements Runnable {
 	private static final String TAG = RequestHandler.class.getSimpleName();
+	private static final int REQUEST_GAP_THRESHOLD = 1000;
 	private final PokemonGo api;
 	private RequestEnvelopeOuterClass.RequestEnvelope.Builder builder;
 	private boolean hasRequests;
-	private List<ServerRequest> serverRequests;
+	private ConcurrentLinkedQueue<ServerRequest> serverRequests;
 	private String apiEndpoint;
 	private OkHttpClient client;
 	private Long requestId = new Random().nextLong();
-
+	private boolean run;
+	private Long lastRequest;
 	private AuthTicketOuterClass.AuthTicket lastAuth;
 
 	/**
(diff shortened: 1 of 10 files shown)


ghost:
@Grover-c13 I tried building it but I got a long log with many error.
Log is in this [async build attempt](http://pastebin.com/VuDGKGv6) Pastebin.

Grover-c13:
thats why its tagged not ready :P its purely for peer review

ghost:
Ah, sorry for the missunderstanding :blush:  I still hope the log is good for something.

pvanassen:
My two cents, If you're going to do TaskUtil.waitForTask, you basically block the thread making the call to get the data. This approach turns your async code attempt back into a synchronous attempt. The only speedup you will achieve is inside the API where multiple calls are made right after each other. 
If you really want to make async work you'll have to either return Futures, Promisses or Observables. If you don't want that, you'll end up working with callbacks everywhere.

Grover-c13:
the methods that use that are basically legacy methods so the API doesnt break once updated, ive been looking at futures but i have some issues with the underlying API design.

pvanassen:
Basically that's the dilemma. Either the API remain backwards compatible or it becomes async. Perhaps it's worth considering having an async API with a wrapper for backwards compatibility?

vmarchaud:
@pvanassen I dont get why you want backwards compatibility since the backwards version will not correctly work without async stuff ? The current lib is high level, users should not have to worry about rate-limit, we dont have any choice to go to async call.

pvanassen:
@vmarchaud I don't want backwards compatibility. That's @Grover-c13 :)
You can change every method to get it async if you'd ask me.

vmarchaud:
BTW i'm starting a lower level lib, only focus on handling network, its not a competitor of this one, but for people who want a lower level, it will be perfect.
You can already watch the repo [over here](https://github.com/vmarchaud/pogoapi) if you are interested.

PS : this lib will be full async/reactive using RxJava.

igoticecream:
Please consider RxJava for async and threads stuff, which handles schedulers and exceptions really well

I feel motivated to work with you vmarchaud

pvanassen:
@igoticecream I totally agree with you but I don't think an extra library is going to make it because of size constraints. If however RxJava is going to be it, I'd happily chip in!

Grover-c13:
the main reason for this PR isnt to really make the API strictly async friendly, its just to make the request handler support scheduling the calls to stop the throttling which needed some async like stuff.

Grover-c13:
and additionally you can sort of do async albeit in a crappy way with it too (not really thread safe async just to note) :P

pvanassen:
@Grover-c13 I'm working on some Future stuff which may allow async stuff and throttling without much API change for the end user, but allows the API user to utilise RxJava if he/she wants. Curious where it will get me.

Grover-c13:
if it works out all well please PR, im really not happy with my current solution

igoticecream:
I think there is no reason to change to async stuff. End users can wrap api calls using rxjava .fromCallable() and get the thing async

pvanassen:
Ok, giving up on Future stuff. Going RxJava on my own fork. Sorry

Grover-c13:
yeah going to need a bit of redesign to work hey :P

pvanassen:
Futures are just horrible. No way of doing callback, no chaining of methods. I end up with tying callback to callback.

Grover-c13:
Think you could make something decent work with RxJava? im okay with using it at the current point.

pvanassen:
Absolutely. I think I'll have a PR tomorrow.

src/main/java/com/pokegoapi/main/RequestHandler.java
```diff
@@ -288,4 +227,27 @@ public void setAltitude(double altitude) {
...
+				} catch (RemoteServerException e) {
+					e.printStackTrace();
+				} catch (LoginFailedException e) {
+					e.printStackTrace();
+				}
+			}
+			try {
+				Thread.sleep(10);
+			} catch (InterruptedException e) {
+				e.printStackTrace();
```
Aphoh, line 249: Use Log.e

Aphoh:
I think the solution here would be an Rx extensions library. It can just be another library that imports this, wrapping the api in Observables and handling rate limiting on it's own.

pvanassen:
You want to stick to Java 7 right?

PR #339, opened 2016-07-29
getAbsoluteMaxCp for pokemonId

ramarro123:
added static version of getAbsoluteMaxCp to be used with pokemonid instead with a pokemon object.

version 2  (lol)


vmarchaud:
After you deleted [this line](https://github.com/ramarro123/PokeGOAPI-Java/blob/ae8fa8b1dcda0d7ee654f979486a296c828cb301/src/main/java/com/pokegoapi/api/pokemon/Pokemon.java#L432), will get merged, thanks for your contribution again ;)

Issue #204, opened 2016-07-25
[Request Feature] Farm Module Pokestops


davidmann4:
**Intro**
The game offers a feature where you can buff a pokestop to lure pokemons. After you used a Module (the item which boosts a pokestop) Pokemons will show up on a regular. Modules are active for 30 minutes.

**Tech**
After you used a Module on a pokestop the pokestop will get a ActivePokemon Value. See a example below.

> disk_encounter
> check to see if the fort has 'lure_info'
> if it has then attempt a disk_enounter
> and if encounter status 1 then try to catch pokemon
> just a heads up not 100% times will you find a successful encounter with lure'd forts mostly you will get it but not always
> working python code:
> https://github.com/j-e-k/pgoapi/blob/master/pgoapi/pgoapi.py#L421

vmarchaud:
Linked to [this issue ](https://github.com/AeonLucid/POGOProtos/pull/103) and duplicate of #151

Issue #151, opened 2016-07-23
CatchablePokemon from fortData


ramarro123:
on CatchablePokemon constructor (with fortData param), the spawnPointId it's set to null.

i think that this constructor will be used to catch pokemon lured on fort, but it won't work
cos enconuter will throw null pointer (spawnpoint = null)

also, if i try to catch without encounter (that i doubt it's the correct way, but just in case)
there is a barrier with if (encountered == false) blablal, so it won't work.

jabbink:
Won't raise a null pointer exception.

Not sure if it works. No lured pokestops around here, hence why the entire code is not called anywhere. Feel free to test it.

ramarro123:
actually, i try it and i got a nullpointer :) 

let me cut & paste exception, give me 2 mins i have lured  poketsops here :D

jabbink:
Oh yeah, I see it, the generated code is fubar

```
      /**
       * <code>optional string spawn_point_id = 2;</code>
       */
      public Builder setSpawnPointId(
          java.lang.String value) {
        if (value == null) {
    throw new NullPointerException();
  }
```

The field is optional (but I don't know if not setting it actually works) but it checks for null...

jabbink:
But if you're near a lure now, quickly try commenting both `.setSpawnPointGuid(getSpawnPointId())` and `.setSpawnPointId(getSpawnPointId())` to see if capturing the pokemon then works, because I fear it doesn't.

ramarro123:
on my way with my laptop :D hehehe

commenting from encounter right?

ramarro123:
another things that make very hard to test this one it's that issue.
https://github.com/Grover-c13/PokeGOAPI-Java/issues/149

i have modify the program to stay in front of the pokestop that i know it's lured, and wait even if it say hasLured == false... wonder why it doesn't report correctly the state of this stop.
....

FIY, after removing the method, encounterPokemon return not success

vmarchaud:
Linked to [this issue](https://github.com/AeonLucid/POGOProtos/pull/103) i guess

Issue #149, opened 2016-07-23
hasLurePokemon() strangeness


ramarro123:
i probably am using it wrong...

i am doing

```
while (true) {
for (stop : go.getMap().getMapObjects().getPokestops())
   stop.hasLurePokemon();

moveABit();
sleep();
}
```

hasLurePokemon actually "blink" (not every 30 mins as expected, but every 2-3 calls)
it say that a pokestop it's in lure mode, then say that it's not in lure mode, and so on, like crazy.

i have try to disable userCache and change the mapExpire, but it doesn't help.

how am i supposed to use it?

Jari27:
Are you sure you are using the right method, i.e. `hasLure()` vs `hasLurePokemon()`? The latter should blink.

ramarro123:
there is no method hasLure in the current checkout

Jari27:
I thought he'd implementend it himself, that's why. It's not weird that hasLurePokemon blinks...

ramarro123:
can you elaborate a bit more why hasLurePokemon it's normal to blink?

i can see that it have an lureexpiresTimeout, but it should expire in 30 min... what i see it's that it blink at every 4000/5000 milliseconds

Jari27:
hasLurePokemon() shows whether there is a pokemon there (due to a lure). I am assuming you capture those, so they dissapear.

ramarro123:
make sense, so it will be necessary, as you said, to add a method to detect if a lure is active (but haven't lured anytihng yet)

it's always better to stay nearby a pokestop that have a lure, and wait till he lure something i think.

it is possible to see if a pokestop has a valid lure plugged on it apart query it via getDetails?

Jari27:
The data is available but I haven't seen it in the api yet.

Edit: actually I just found `pokestop.getFortData().getLureInfo().` Maybe fiddle around with this?

ramarro123:
pokestop.getFortData().getLureInfo()
return the same data used in hasLurePokemon, so the results will be the same (blinking)

Pyr0x1:
I am checking if a pokestop has a lure module on it in this way and it seems working, even if I don't know what a TROY_DISK is:

```
Collection<Pokestop> pokeStops = go.getMap().getMapObjects().getPokestops();
            for (Pokestop stop : pokeStops) {
                if (stop.getId().equals(Preferences.pokestopId)) {
                    Log.i(TAG, "Found pokestop");
[...]

PR #344, opened 2016-07-29
catch pokemon lured by disk on pokestop

ramarro123:
adding the ability to catch a pokemon lured by disk on fort.

had to refactor EncounterResult in a way that i don't like much but i don't want to break too much the compatibility with the previous version (that's why i convert DiskRes to EncRes)

*implemented the call DISK_ENCOUNTER that have to be used instead of ENCOUNTER

the relative issues are:
#151 #149 #204

now, let's see if this time i have mess with git again ;)

### DIFF
--- a/src/main/java/com/pokegoapi/api/map/pokemon/CatchablePokemon.java
+++ b/src/main/java/com/pokegoapi/api/map/pokemon/CatchablePokemon.java
@@ -21,11 +21,13 @@ import POGOProtos.Map.Fort.FortDataOuterClass.FortData;
 import POGOProtos.Map.Pokemon.MapPokemonOuterClass.MapPokemon;
 import POGOProtos.Map.Pokemon.WildPokemonOuterClass.WildPokemon;
 import POGOProtos.Networking.Requests.Messages.CatchPokemonMessageOuterClass.CatchPokemonMessage;
+import POGOProtos.Networking.Requests.Messages.DiskEncounterMessageOuterClass;
 import POGOProtos.Networking.Requests.Messages.EncounterMessageOuterClass;
 import POGOProtos.Networking.Requests.Messages.UseItemCaptureMessageOuterClass;
 import POGOProtos.Networking.Requests.Messages.UseItemCaptureMessageOuterClass.UseItemCaptureMessage;
 import POGOProtos.Networking.Requests.RequestTypeOuterClass;
 import POGOProtos.Networking.Responses.CatchPokemonResponseOuterClass.CatchPokemonResponse;
+import POGOProtos.Networking.Responses.DiskEncounterResponseOuterClass;
 import POGOProtos.Networking.Responses.EncounterResponseOuterClass;
 import POGOProtos.Networking.Responses.EncounterResponseOuterClass.EncounterResponse;
 import POGOProtos.Networking.Responses.UseItemCaptureResponseOuterClass;
@@ -48,8 +50,15 @@ public class CatchablePokemon {
 	private static final String TAG = CatchablePokemon.class.getSimpleName();
 	private final PokemonGo api;
 
+	private enum EncounterKind {
+		normal,
+		disk
+	}
+
+
 	@Getter
 	private final String spawnPointId;
+
 	@Getter
 	private final long encounterId;
 	@Getter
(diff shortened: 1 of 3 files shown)


ramarro123:
as a follow up, it's not that easy to step into a pokestop with an active lured pokemon :)

i have finish my lure so i can't test anymore.

that's the result
catturato EEVEE di tipo: disk

tipo: disk
on my log means that the encounterKind = disk

if someone can do a bit of test to confirm, i appreciate

igoticecream:
Lol, i was going to submit this exact pull request. Yes it works, you can construct a catchablepokemon with fortdata and set spawn point it as getFortId!!!

Check out: https://github.com/AeonLucid/POGOProtos/pull/103

ramarro123:
:)

JamesPeters98:
This has been working perfect on a few lures near me.

src/main/java/com/pokegoapi/api/map/pokemon/CatchablePokemon.java
```diff
@@ -99,42 +108,50 @@ public CatchablePokemon(PokemonGo api, WildPokemon proto) {
 		this.expirationTimestampMs = proto.getTimeTillHiddenMs();
 		this.latitude = proto.getLatitude();
 		this.longitude = proto.getLongitude();
+		this.encounterKind = EncounterKind.normal;
 	}
 
 	/**
 	 * Instantiates a new Catchable pokemon.
```
CaptnBlubber, line 115: Would be clearer if you stated here that this is from a Lured Fort

src/main/java/com/pokegoapi/api/map/pokemon/CatchablePokemon.java
```diff
@@ -151,18 +168,38 @@ public EncounterResult encounterPokemon() throws LoginFailedException,
...
+				.setPlayerLongitude(api.getLongitude())
+				.setFortId(getSpawnPointId()).build();
+		ServerRequest serverRequest = new ServerRequest(
+				RequestTypeOuterClass.RequestType.DISK_ENCOUNTER, reqMsg);
+		api.getRequestHandler().sendServerRequests(serverRequest);
+		DiskEncounterResponseOuterClass.DiskEncounterResponse response = null;
+		try {
+			response = DiskEncounterResponseOuterClass.DiskEncounterResponse
+					.parseFrom(serverRequest.getData());
+		} catch (InvalidProtocolBufferException e) {
```
CaptnBlubber, line 189: Why do you use try catch here when the rest of this method already has the a throws Signature for LoginFailedException and RemoteServerException ?

src/main/java/com/pokegoapi/api/map/pokemon/EncounterResult.java
```diff
@@ -17,38 +17,108 @@
...
+	 * Return the status of the encounter
+	 *
+	 * @return status of results
+	 */
+	public abstract EncounterResponse.Status getStatus();
 
-	public boolean wasSuccessful() {
-		return response != null && getStatus() != null && getStatus().equals(EncounterResponse.Status.ENCOUNTER_SUCCESS);
-	}
+	public static class DiskEncounterResult extends EncounterResult {
```
CaptnBlubber, line 37: Not a big fan of Inner Classes, it makes Class Files large (and therefore harder to read/understand/debug) and Code instantiations of Objects uglier.
Personaly I would Have a Seperate file for Bot DiskEncounterResult.java and EncounterResult.java

src/main/java/com/pokegoapi/api/map/pokemon/CatchablePokemon.java
ramarro123, line 189: just for consistency with the rest of API.

also i catch InvalidProtocolBufferException and throw RemoteServerException

take a look to the rest of api, it's done in the same way

src/main/java/com/pokegoapi/api/map/pokemon/EncounterResult.java
ramarro123, line 37: not a big fan of inner as well, but in this case the whole class it's 124 line (including the 2 inners), it won't be too hard to navigate or read that.

if there will be need for other EncounterResponse (tm) i will split it in different classes

ryangardner:
Is the only holdup around the style issue regarding the inner classes? 

I started down a very similar path today to try to get the lured pokemon working - just enabling it you can see a _tremendous_ different in how many pokemon are in a densely lured area. 

If there is any way I can help out I'm willing to do so. This is a really important change to get in

Grover-c13:
Iwas holding off til we had the server rate limit fixed. Needs to have conflicts resolved now, so ill just rework it into a new PR

Grover-c13:
Redone in PR #418, ty for contribution @ramarro123

PR #350, opened 2016-07-29
Changed CredentialProviders to allow easier extending of existing CredentialProviders

s7092910:
**Changes made:**
- Changed private fields in GoogleUserCredentialProvider and PTCCredentialProvider to protected so they are visible in child classes.
- Allows easier modifications of these classes outside of the API.


Jari27:
Right now there are only PTC and Google logins. I don't see that changing in the near feature. If it doesn't, less visibility is better, imho.

Thoughts?

vmarchaud:
If you want to make change to a provider, why not implementing your own since the interface is public ? It should be simpler for you i think.

s7092910:
@vmarchaud that is what I told the rest of my team, but they said it would be better to extend the already existing CredentialProvider. And they wanted me to make this PR. But I would say this would be a nice idea, for instance I know people have been asking for a method in the PTCCredentialProvider to take in a TokenId. This in theory would allow them to do that along with if they want to change a bit of how the internals of the Provider works, without having to rewrite everything.

vmarchaud:
@s7092910 
I agree that the lib is actually may be too high level for people that know who the networking is done, thats why i'm making [another lib](https://github.com/vmarchaud/pogoapi) (that maybe will get introduced into the java implementation) that doesnt force any implementation of auth, just provide interface to do it.
In regard of your PR, wait for @Grover-c13 for this PR, i personally prefer that your team make their provider has they want to over allow the modification of our provider.

rajulbhatnagar:
If you want to implement your provider to suit your needs you can do so by implementing the interface however I do feel that making those fields protected is reasonable enough. I dont see any downside to it. @vmarchaud Do you see any downsides to this?

Grover-c13:
@s7092910 are you on the PKRE slack?

s7092910:
@Grover-c13 No I'm not, do you want me to join?

Grover-c13:
nothing major just wanted to discuss something relating to the login providers and the API :P

PR #355, opened 2016-07-30
Fixed Pokestop.hasLurePokemon when returning false with an active lure on it

s7092910:
**Changes made:**
- Pokestop.hasLurePokemon will now properly return true if the Pokestop has a Lure Pokemon on it and the Lure Module is active.


AbandonedCart:
Thank you for saving me a huge amount of stress. I've been trying to figure out why every list of nearby lures was empty and thinking I had to go put up more for testing.

This is deprecated, but the new method has not had any better results.

PR #357, opened 2016-07-30
[DON'T MERGE] Initial Rxjava Async [WIP]

mashurex:
**Changes made:**
- Added RxJava library
- Changed RequestHandler to use deferred Observables
- Added some last request time checking to make sure requests don't get throttled
- Added some retries when `InvalidProtoBuffer` errors occur
- Changed `PlayerProfile` and `Inventories` classes to utilize the new Observables
- Added `synchronized` methods to PokemonGO, PlayerProfile, and Inventories classes for thread safety
- Merged the latest Development branch changes into my branch


Danm72:
So awesome. GG @mashurex.

mashurex:
There's a lot more RxJava work that can be done to make it really hum. I didn't subscribe on any particular schedulers or anything and it can be used to manage the cache- which it's currently not. It's a very initial and naive implementation at this point.

FabianTerhorst:
In my opinion using rxjava for the async work is almost the best way. It should be totally ok to add rxjava to the project.

src/main/java/com/pokegoapi/api/PokemonGo.java
```diff
@@ -26,34 +26,19 @@
 import com.pokegoapi.main.RequestHandler;
 import com.pokegoapi.util.SystemTimeImpl;
 import com.pokegoapi.util.Time;
-import lombok.Getter;
-import lombok.Setter;
```
FabianTerhorst, line 30: Why you remove lombok from this file?

mashurex, line 30: @FabianTerhorst my IDE probably auto-removed it because I HAD made all the get/set synchronized. As I look at them now it seems that I lose them in one of my changes and need to add  synchronized back in.

FabianTerhorst:
some of the added docs has a different number of spaces, is checkstyle working?

src/main/java/com/pokegoapi/google/common/geometry/S2EdgeIndex.java
```diff
@@ -529,7 +529,7 @@ private void getEdgesInChildrenCells(S2Point a, S2Point b, List<S2CellId> cover,
     }
   }
 
-  /*
+  /**
```
FabianTerhorst, line 532: should be against checkstyle?

FabianTerhorst:
pokenames is already merged in dev branch, why is it still in this pr?

mashurex:
I'll clean it up, I think I screwed up my tracking to keep this up to date with Development at the same time.

src/main/java/com/pokegoapi/google/common/geometry/S2EdgeIndex.java
mashurex, line 532: I ran everything against check style, spaces look the same but I added an extra \* to make it a docblock comment

PR #360, opened 2016-07-30
Async implementation based in Futures

pvanassen:
I started of with Observables but without Java 8 it requires a load of boilerplating. The same boilerplating I need for futures. So I went forward with those. This means, no extra library needed. 

I started of with maps and I'm able to do all the async work transparently. It's up to the API user to use the Future method, or the old one. In the background all API calls are limited to 3 per second. If you use the async calls you get a lot more performance. If not, it just works.


Grover-c13:
What are your thoughts on PR #357 pvanassen?

pvanassen:
@Grover-c13 Can't really tell since most of the files change are not related to Observables but to JavaDoc or other minor changes. 
The request handling definitly uses Observables, and in a propper way it seems. Curious where it will go to.

pvanassen:
I'm doing a test run with all modified code now so I can fix everything that is broken.

FabianTerhorst:
@Grover-c13 I prefer RxJava more then the default Java implementation with future objects. With rxjava you can simplify many processes that are currently a little bit complex, like the error handling.

pvanassen:
Current state of this branch is, it can be merged. All current calls are handled through the throttle. All existing calls still work. If you have a multi-threaded environment, you can fire off multiple requests. They will be bundled before being send to Niantec. 
Some calls have been converted to expose a Future. They can easily be wrapped into an Observable by the implementing party. 

Most work to be done is to convert more calls to be async.

pvanassen:
@FabianTerhorst I agree. But if you want the highest usage of your API, you'd better have a good transition towards that. Also if slf4j with its 41KB is too much of a size increase, RxJava with 1.1MB definitly is an issue. This implementation allows you to use it with Observables if you want.

pvanassen:
Regarding testing, I am running my own code against it without modifications. I've had one or two issues with empty requests, so I've upped the timer a bit. Apart from that it works like a charm

Grover-c13:
Ill accept this then until rx is ready.

Issue #250, opened 2016-07-26
WildPokemon getTimeTillHiddenMs() returns large negative numbers


MrPat:
**Description:**
Sometimes the getTimeTillHiddenMs() method returns a feasible number, but a lot of times it returns a huge negative number which obviously isn't correct. Could this be just an overflow of the int32 that holds it? If so is there a way to infer the actual value based on MAX_INT?

**Example of a WildPokemon toString()**
encounter_id: 17736830237155595229
last_modified_timestamp_ms: 1469509425575
latitude: xx.15776268002814
longitude: -xx.74958767808612
spawn_point_id: "xxxxxa87fef"
pokemon_data {
        pokemon_id: CATERPIE
}
time_till_hidden_ms: -630610344

(x's are just to obfuscate my actual location)

vmarchaud:
Its an unsigned long, there arent handled natively by Java, you need to use java 8 methods : `Long.toUnsignedString(var)` to print them into your console.
Note that these methods doesnt exist in Java 7 or under, you need to figurate a way to print them.

MrPat:
Ok I've been working on converting the returned value to an unsigned value. Now instead of something like -701128148 I get 3593839148, which is obviously not the amount of ms left. This number does decrease perfectly with time though. The digit in the thousands position decreases with seconds. I can't figure out why it corresponds to time because there definitely aren't 3593839 seconds until the despawn. Does anybody know what to do with this number after converting it to a Java friendly value? It doesn't seem to be in ms even though it is labeled as such and behaves like a time value.

Grover-c13:
if you bitshift the time 16 bits right (3593839148 >> 16) do you get the number you are expecting?

MrPat:
Thanks for the response. I just tried bitshifting a time (3590370382) and it gave me 54784, which looked promising but it doesn't decrease at all with seconds. I must be doing the unsigned conversion wrong.

Say the getTimeTillHiddenMs() method returns -704596914. I'm currently ANDing that result with 0xffffffffL but that obviously isn't the right thing to do. Specifically I'm doing this `getTimeTillHiddenMs() & 0xffffffffL`

I've also tried the following but it gave a huge number that I couldn't do anything with. I should note that this number also decreased properly with time.

```
public static String asString(long l) {
        return l >= 0 ? String.valueOf(l) : toBigInteger(l).toString();
}

public static BigInteger toBigInteger(long l) {
[...]

PR #372, opened 2016-07-30
Fix expiration time issues related to Wild Pokemon

broach:
**Fixed issue:** [Reference the issue number here, or remove if not a fix]
#250 WildPokemon getTimeTillHiddenMs() returns large negative numbers

**Changes made:**

Wild pokemon have two fields in the protobuf;
LastModifiedTimestampMs and TimeTillHiddenMs. (The
TimeTillHiddenMs will often be a negative number, see below).

Example:
LastModifiedTimestampMs: 1469902449401
TimeTillHiddenMs: -1023634170

**(edit)** I didn't complete this thought, sorry. 
To get the actual expire time, you need to add them together.

I'm not entirely sure why occasionally TimeTillHiddenMs is a large negative number, resulting in a time in the past. I need to do some more research and see if the pokemon itself is actually spawned where it says it is even though the expiration is in the past.

In addition, you will also see that in a single query you get back
the same Pokemon (same encounter id) in both Wild and Catchable.
The catchable will have a -1 expiration time if the above happens.

I've modified the CatchablePokemon constructor to calculate the
correct expiration time for WildPokemon and also screen out the
-1 expiration Catchables in MapObjects.addCatchablePokemons()
(They're going to be in the Wild array with the correct expiration)

### DIFF
--- a/src/main/java/com/pokegoapi/api/map/MapObjects.java
+++ b/src/main/java/com/pokegoapi/api/map/MapObjects.java
@@ -82,7 +82,11 @@ public class MapObjects {
 			return;
 		}
 		complete = true;
-		this.catchablePokemons.addAll(catchablePokemons);
+		for (MapPokemon p : catchablePokemons) {
+			if (p.getExpirationTimestampMs() != -1 ) {
+				this.catchablePokemons.add(p);
+			}
+		}
 	}
 
 	/**
--- a/src/main/java/com/pokegoapi/api/map/MapObjects.java
+++ b/src/main/java/com/pokegoapi/api/map/MapObjects.java
@@ -82,11 +82,7 @@ public class MapObjects {
 			return;
 		}
 		complete = true;
-		for (MapPokemon p : catchablePokemons) {
-			if (p.getExpirationTimestampMs() != -1 ) {
-				this.catchablePokemons.add(p);
-			}
-		}
+		this.catchablePokemons.addAll(catchablePokemons);
 	}
 
 	/**
--- a/src/main/java/com/pokegoapi/api/map/pokemon/CatchablePokemon.java
+++ b/src/main/java/com/pokegoapi/api/map/pokemon/CatchablePokemon.java
@@ -99,7 +99,7 @@ public class CatchablePokemon {
 		this.spawnPointId = proto.getSpawnPointId();
 		this.encounterId = proto.getEncounterId();
 		this.pokemonId = proto.getPokemonData().getPokemonId();
-		this.expirationTimestampMs = proto.getTimeTillHiddenMs();
+		this.expirationTimestampMs = proto.getLastModifiedTimestampMs() + proto.getTimeTillHiddenMs();
 		this.latitude = proto.getLatitude();
 		this.longitude = proto.getLongitude();
 	}


broach:
Researched / worked a bit more on this. You get a negative value when the Pokemon is at a spawn point that lasts 30 minutes. For the first 15 minutes you get negative values, then for the second 15 minutes you get positive, correct values.

The negative value is time _based_, and it's in milliseconds. If you compare  a value against a previous one and note the times you retrieved them you'll see that the result is actually the number of milliseconds elapsed. 

Unfortunately, it's not counting down to zero, nor is it a constant. The range of values used is different every time. I wracked my brain trying to figure out some time based value the the negative values correlate to and came up with nothing.

AbandonedCart:
According to your notes, you intend to `screen out the -1 expiration Catchables in MapObjects.addCatchablePokemons() (They're going to be in Wild array with the correct expiration)`

Won't this mean that for anyone using strictly catchable to return the number of "useful" map points available, that list would be missing all of the ones that needed an adjusted expiration?

Grover-c13:
its not actually a negative value, its an unsigned long being placed into an int32 and overflowing

AbandonedCart:
It seems like that would be fixed simply by returning a proper cast of the number.
Does that resolve the issue?

broach:
@Grover-c13 That makes no sense. The value is the number of milliseconds before the Pokemon disappears; when the value is positive it's not a timestamp.  Even if that were 30 minutes (1800000 milliseconds) it would happily fit in an signed int (max 2147483647) so I don't see how it's an overflow.

@LoungeKatt Re: leaving them out of catchable; Yeah, I went back and forth on that. Just leaving them in catchable is probably better given your comment. As for casting that number ... no? Even if it were an overflow a cast to a long would simply be a widening conversion and you'd wind up where you started.

Grover-c13:
Yeah i know it doesnt make sense, but thats what the server sends sometimes. You can see it in raw data dumps (eg: 
11: 18446744072919745186)

broach:
OK, which ... that's an equally useless number :) I'm guessing it's a bug on their side. I have no idea what the actual game is doing with it, but it's possible it doesn't care about the disappear time so no harm no foul. 

Either way, the change I made does give you the correct expire when that number isn't busted. (the current code sets the expire to TimeTillHiddenMs which is wrong in all cases). I'll revert the bit that excludes the -1 from catachable.

AbandonedCart:
@broach Why would you cast an overflowed int32 to a long? You cast it to long in the initialization. Perhaps this is becoming so complicated because you haven't gone far enough back in the process.

Grover-c13:
Mm its not entirely useless, ulong max: 18446744073709551615,
18446744073709551615-18446744072919745186=789807104, one week expiry times :P?

Grover-c13:
Could anyone get me some lat/longs of the spawn points with weird expiry times.

Grover-c13:
Also just to add, after investigating some more -1 on MapPokemon, its appearing as a long (2^64, or -1 cast to ulong), protos to compare:

> spawn_point_id: \"2a32afed0a5\"
> encounter_id: 16191425754593857709
> pokemon_id: PIDGEOTTO
> expiration_timestamp_ms: -1
> latitude: -31.90713105143732
> longitude: 115.8313344267818
> 
> ```
>   1: "2a32afed0a5"
>   2: 0xe0b37ff01dad90ad
>   3: 17
>   4: 18446744073709551615 // the -1
>   5: 0xc03fe839bd971c01
>   6: 0x405cf534954fc445
> ```

and raw proto to reference field numbers:

> message MapPokemon {
>   string spawn_point_id = 1;
>   fixed64 encounter_id = 2;
>   .POGOProtos.Enums.PokemonId pokemon_id = 3;
>   // After this timestamp, the pokemon will be gone.
>   int64 expiration_timestamp_ms = 4;
>   double latitude = 5;
>   double longitude = 6;
> }

The pokemon is 100% catchable and shows on the client, however it just kind of popped up after a few moments hanging around (only tested once), not sure if this is common.

So in conclusion not really any closer to working this out.

Grover-c13:
Just to add, when it did appear it was given a valid expiry of 879632

broach:
I'll give you some lat/lng of some of these guys in a bit here.

The deal is they will show up and be catchable (in the game) while reporting that large / overflowed / whatever number in TimeTillHiddenMs. That number counts up (toward 0), and if you compare it over time it _is_ counting in milliseconds (e.g. get the poke at 16:10:00 and 16:15:00, subtract the second number from the first and you'll see that the difference is 300000). The thing is, it's not the same range every time. If it was, it would be easy to deal with (e.g. if it always counted to say, -1037182165 before flipping to positive, that'd be simple). 

Usually at some point, that number changes to positive and makes sense. Every one I've seen, as you also noticed, has a 15 minute clock on it when that happens (just short of 900000 depending on when your query lands). In rare cases it _never_ goes positive and the little bugger just disappears.

broach:
Here's one right now, just starting my scanner (native android app I've been playing with). Ignore the 0 expiration_time - I set that to distinguish them in my app and update it if TimeTillHiddenMs goes positive). 

07-31 12:09:28.444 3322-3462/net.mostlyharmless.pokemontracker D/Scan: Fetched 1 Wild Pokemons
07-31 12:09:28.444 3322-3462/net.mostlyharmless.pokemontracker D/Scan: Wild Pokemon: 

{ latitude=39.98415182733396 
  expiration_time=0 
  insert_timestamp=1469988568444 
  pokemon_name=Rattata 
  encounter_id=8433726560817769565 
  longitude=-104.77921413451496 
  pokemon_id=19 } 
  LastModifiedTimestampMs: 1469988568808 
  TimeTillHiddenMs: -1109753577

broach:
And ... he just flipped to positive. 

07-31 12:16:57.665 3322-3462/net.mostlyharmless.pokemontracker D/Scan: Fetched 1 Wild Pokemons
07-31 12:16:57.665 3322-3462/net.mostlyharmless.pokemontracker D/Scan: Wild Pokemon:

{ latitude=39.98415182733396 
  expiration_time=1469989858460 
  insert_timestamp=1469989017665 
  pokemon_name=Rattata 
  encounter_id=8433726560817769565 
  longitude=-104.77921413451496 
  pokemon_id=19} 
  LastModifiedTimestampMs: 1469989018030 
  TimeTillHiddenMs: 840430

AbandonedCart:
Unless an exact time can be extracted, it seems more effective to handle this in the usage end. 

The last modified time is the time of the last "ping" for the particular object, but adding that to the negative only results in `the last time it was modified minus the overflow value of an int32`. 
More or less it is just an imaginary number.

The alternative would be to do something like `wait for this to become positive, then set a definitive timeout. Until such time, check the current time until hidden`

PR #376, opened 2016-07-30
Implemented inventory settings

pvanassen:
Implemented inventory settings


src/main/java/com/pokegoapi/api/settings/InventorySettings.java
```diff
@@ -1,13 +1,27 @@
 package com.pokegoapi.api.settings;
 
-import POGOProtos.Settings.InventorySettingsOuterClass;
```
Aphoh, line 3: Why get rid of the outer class import?

pvanassen, line 3: Because it was unused. The update method already used 'POGOProtos.Settings.InventorySettingsOuterClass.InventorySettings'. I did not change that part and therefor my IDE notified me the import was not used.

PR #389, opened 2016-07-31
Optimize imports

Cyperghost:
Optimize imports in each files


Grover-c13:
Could you please rebase

PR #400, opened 2016-07-31
Add evolution auxiliaries

svarzee:
**Changes made:**

Added some useful evolution info utils


src/main/java/com/pokegoapi/api/pokemon/EvolutionForm.java
```diff
@@ -0,0 +1,29 @@
+package com.pokegoapi.api.pokemon;
```
FabianTerhorst, line 1: Add license header

PR #420, opened 2016-08-01
Master

Grover-c13:
0.3


FabianTerhorst:
when you released 0.3 i will create the new branch to split the modules. I could also add a maven upload script so that everybody is using the same provider.

PR #421, opened 2016-08-01
Add catch with razzberry and specified pokeball

AbandonedCart:
Convenience method to support pokeball selection when using razzberries


AbandonedCart:
Closing this in favor of one built on a proprietary fork branch

PR #422, opened 2016-08-01
Catch pokemon with the current available best ball

Cyperghost:


FabianTerhorst:
Maybe add an parameter to make the masterball excludable, because the masterball is very rare.

src/resources/protobuf
```diff
@@ -1 +1 @@
-Subproject commit 333e54707ca221324c6b9eedb6a7bc2fcae02270
+Subproject commit 636c52f4df519dd74000bbe8e42b8cb32ba8b51a
```
FabianTerhorst: you updated the protos?

Cyperghost: Yes,there is a problem?

FabianTerhorst: because #419 used a newer version

Cyperghost: Oh, thanks.
I forgot to change the branch ;(

FabianTerhorst:
👍

PR #435, opened 2016-08-01
remove unnecessary catchablePokemons cache - use map objects cache

mgorski-mg:
Catchable Pokemons can not be refreshed if different method update map objects cache

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/map/Map.java
+++ b/library/src/main/java/com/pokegoapi/api/map/Map.java
@@ -71,7 +71,6 @@ public class Map {
 	private static int RESEND_REQUEST = 5000;
 	private final PokemonGo api;
 	private MapObjects cachedMapObjects;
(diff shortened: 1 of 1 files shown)


Grover-c13:
It is necessary, consider a situation where you get catchable pokemon, catch one, and then get catchable pokemon again. You will have a list with the caught pokemon and when you try to encounter it will get an error. I will rework it a bit so that it is cleared when other getMapObjects call

mgorski-mg:
You're right but if we don't fix it sometimes catchable pokemons can always be 0.
You're issue needs changes in CatchablePokemon class in catchPokemonAsync - line 250
`CatchablePokemon.this.api.getMap().getCatchablePokemon().remove(this);`
we should update cachedMapObjects in Map class instead of catchablePokemons.

I can add this and reopen this issue if you accept this solution :)

mgorski-mg:
Ok your new commit fix it. Thanks!

PR #436, opened 2016-08-01
add binary plugin to library module

FabianTerhorst:
use ./gradlew :library:bintrayUpload to test script (remove org.gradle.configureondemand=true in grade properties to resolve jar generation error)

Todo: javadoc isnt submitted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,19 @@ ___
 # How to import
 
   Import from Maven/Gradle/SBT/Leiningen using JitPack : [![](https://jitpack.io/v/Grover-c13/PokeGOAPI-Java.svg)](https://jitpack.io/#Grover-c13/PokeGOAPI-Java)
-  
+
+  ```groovy
+  allprojects {
+    repositories {
+        jcenter()
+    }
+  }
+
+  dependencies {
+    compile 'com.pokegoapi:library:0.3'
+  }
+  ```
+
   After you clicked on this link, jitpack will show you multiple build (try use the latest one since the api grow everyday).
   JitPack will show an example for each dependency manager to include our API into your project.
 
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -1,9 +1,11 @@
 buildscript {
     repositories {
         mavenCentral()
+        jcenter()
     }
     dependencies {
         classpath 'com.google.protobuf:protobuf-gradle-plugin:0.7.7'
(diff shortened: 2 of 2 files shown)


FabianTerhorst:
@Grover-c13 you have to create a account at bintray.com . Then you have to change your gradle properties and add there the user and apiKey, but don´t commit the properites. After the first upload script you have to confirm the project on your account and add a request for publish to jcenter.

AbandonedCart:
What is the benefit of this over, for example, Github releases (which are already supported and require no additional imports to be added)?

s7092910:
@LoungeKatt you just do a gradle build and it puts it on bintray which is jcenter. So anyone wanting to add it to their gradle build, they can just do 'compile 'project-blah:version-blah'. And depending on your IDE, it will notify you if there is a new version out. A lot easier on the user to use. Jcenter is sort of the new standard when it comes to Java libraries or even Android libraries.

FabianTerhorst:
It's much Cleaner to use then jitpack. The jcenter releases will never be end unresolved dependencies and you know exactly what is released.

AbandonedCart:
@s7092910 Not sure I would call one less line a whole lot easier, but thanks for the explanation anyway. With how often the repo updates, I assume this is more of a long term goal.

FabianTerhorst:
Normally this is the first think thats happening before a library is getting released.

FabianTerhorst:
he is failing, because local properties doesn´t exists in github and shouldn´t exists.

Grover-c13:
wont it fail for all builds in future then?

FabianTerhorst:
i will fix it. Maybe submit a empty local.properties file?

FabianTerhorst:
this commit could fix it.

FabianTerhorst:
@Grover-c13 travis is fixed

FabianTerhorst:
you created an account?

FabianTerhorst:
i have tested the upload https://bintray.com/fabianterhorst/maven/PokeGOAPI

Grover-c13:
yeah setting it up now, please feel free to join slack to offer me technical support :P

LoungeKatt:
@FabianTerhorst When you update the README, please make sure to update the **ENTIRE** relevant section, not just the single line of code that changed. You left a whole section on using jitpack, including the link to jitpack, wrapping this new method.

PR #442, opened 2016-08-02
login

Grover-c13:
Fix logins not refreshing after 1.5 hours, remove print from catchable pokemon, add checks for valid long/lat, throws illegal state exception if attempting to get map with out setting location at least once.

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/PokemonGo.java
+++ b/library/src/main/java/com/pokegoapi/api/PokemonGo.java
@@ -38,22 +38,18 @@ public class PokemonGo {
 	@Getter
 	RequestHandler requestHandler;
 	@Getter
-	Map map;
-	@Getter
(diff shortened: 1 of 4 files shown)


library/src/main/java/com/pokegoapi/api/PokemonGo.java
```diff
@@ -151,4 +146,43 @@ public Settings getSettings() throws LoginFailedException, RemoteServerException
...
+		latitude = value;
+	}
+
+	/**
+	 * Validates and sets a given longitude value
+	 *
+	 * @throws IllegalArgumentException if value exceeds +-180
+	 */
+	public void setLongitude(double value) {
+		if (value > 1800 || value < -180) {
```
jabbink, line 170: > 1800?

PR #447, opened 2016-08-02
Catch pokemon automatically poke ball

Cyperghost:
This function choose automatically the right ball to catch a pokemon

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/PokemonGo.java
+++ b/library/src/main/java/com/pokegoapi/api/PokemonGo.java
@@ -152,7 +152,7 @@ public class PokemonGo {
 	/**
 	 * Validates and sets a given latitude value
 	 *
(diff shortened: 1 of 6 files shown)


FabianTerhorst:
you run checkstyle?

Cyperghost:
Yes, and no errors ;)

pvanassen:
What I've seen is the probability levels overall being very low. Have you tested this code in the wild? It might in most cases select the best pokeball you have. Personally I use a cutoff of 35% because 50% was too high, let alone 75%.

Cyperghost:
I often test it with 30% than the pokemons flee.
I think i can changed it to 50%, but less i think it´s that flee,too.

pvanassen:
It would be interesting to gather some data. I don't comment on this PR because you need to change it btw. More as a point of debate.

john086:
Where is this method ? Is an update on catch pokemon method?

PR #448, opened 2016-08-02
update dependencies and cleanup gradle files

FabianTerhorst:
update okhttp to a stable version
update lombok

PR #453, opened 2016-08-02
Fix README, Compile the output expected

AbandonedCart:
This moves the bintrary instructions as the primary method, but
continues to list the alternatives. More importantly, the user will not
mistakenly follow both sets of instructions as a single process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,8 +27,6 @@ ___
 
 # How to import
 
-  Import from Maven/Gradle/SBT/Leiningen using JitPack : [![](https://jitpack.io/v/Grover-c13/PokeGOAPI-Java.svg)](https://jitpack.io/#Grover-c13/PokeGOAPI-Java)
-
   ```groovy
   allprojects {
     repositories {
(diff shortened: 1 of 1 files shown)


FabianTerhorst:
can you replace compile 'com.pokegoapi:library:0.3' with compile 'com.pokegoapi:PokeGOAPI-library:0.3.2'

AbandonedCart:
@FabianTerhorst I did not change the specific code when submitting the PR. I only moved things around so it wasn't injected into the middle of other instructions.

I am guessing this was the final build after setting up the service, so it was submitted as a separate commit in case it needs to be dropped / edited later.

AbandonedCart:
@FabianTerhorst @Grover-c13 is the base library what is being hosted on bintray? As it stands now, using the information provided, that is what you have as the published dependency.

I have tried to clean it up as best I can, but it is important to watch for overlapping definitions and matching what you intend to get with what you are actually building.

Danm72:
This is going to majorly throw people off.

LoungeKatt:
The entire implementation was too much, too soon.
Something like this is best done either from the beginning, or after things have reached a stable point.

It can't be helped now, though. The changes here are only fixing all the undocumented, mismatched changes that have already been merged.

build.gradle
```diff
@@ -1,5 +1,5 @@
 group = 'com.github.aphoh'
```
FabianTerhorst, line 1: also change to com.pokegoapi ?

AbandonedCart, line 1: It doesn't make a difference, but isn't this stuff you should have done in the original PR?

LoungeKatt:
After running a bunch of tests and trying various changes, something in the recent changes has broken local compile. The only reason it wasn't detected before was the copy script I have ignoring the new library because the name wasn't updated to match what it was building.

build.gradle
FabianTerhorst, line 1: i don´t think this variable is needed in this file anyway.

FabianTerhorst:
can you remove the changes from the gradle files?

FabianTerhorst:
https://github.com/Grover-c13/PokeGOAPI-Java/pull/448/commits/4bd9f2aeffe69d78e1e063fa698dcec4101b2f67

LoungeKatt:
@FabianTerhorst No. That is an unmerged PR. As I stated when you originally requested to change the version number, I did not want to add that because it was too volatile.

Get your thoughts together, make a final commit that is actually complete, and then I will update the PR to reflect it. Right now, you have another mismatched, disorganized mess.

AbandonedCart:
This now depends on https://github.com/Grover-c13/PokeGOAPI-Java/pull/459 to handle the versioning in the gradle build files and only updates the documentation to match.

PR #454, opened 2016-08-02
Add catch with razzberry and specified pokeball

AbandonedCart:
Convenience method to support pokeball selection when using razzberries


FabianTerhorst:
can you replace ompile 'com.pokegoapi:library:0.3' with compile 'com.pokegoapi:PokeGOAPI-library:0.3.2'

LoungeKatt:
@FabianTerhorst Wrong PR

PR #458, opened 2016-08-03
update dependencies and cleanup gradle files

FabianTerhorst:
update dependencies
replace jetpack dependence with bintray upload
remove mavenCentral
use http://semver.org/ for version numbers

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -1,8 +1,5 @@
-version = '0.0.1-SNAPSHOT'
-
 buildscript {
     repositories {
-        mavenCentral()
         jcenter()
     }
     dependencies {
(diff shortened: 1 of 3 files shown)


AbandonedCart:
Is semver just a GUI interface that tells you the version number or is the implementation of doing that going to be a separate PR?

FabianTerhorst:
okhttp 3.4.1 is more stable to use so i updated it to 3.4.1. Its also the same version the google auth lib is using. For lombok the plugin will notify when the version is outdated.

library/build.gradle
```diff
@@ -37,6 +24,8 @@ sourceSets {
     }
 }
 
+archivesBaseName = 'PokeGOAPI-library'
```
AbandonedCart, line 29: One way to allow more control from the master build script would be `archivesBaseName = archivesBaseName + '-library'` where the master build script has `archivesBaseName = 'PokeGOAPI'`

FabianTerhorst:
#459

AbandonedCart:
You noted `use http://semver.org/ for version numbers` but all I saw at the site was a thesis on how to choose version numbers, so is there more to it or was it just a reference?

FabianTerhorst:
vi is my friend for squashes^^

FabianTerhorst:
Its a reference for the scheme that should be used now

FabianTerhorst:
every bugfix +1 on the last number and so on.

FabianTerhorst:
every bugfix release*

AbandonedCart:
I was aware how it works, just wasn't sure why it was a commit note. I thought it might be a service to handle incrementation. That's SOP.

PR #475, opened 2016-08-03
fix getting pokemons from pokestop with lure

mgorski-mg:
Remove checking EncounterId while getting catchable pokemons from pokestop with lure.

I'm also thinking about making bigger range thank 30 on lured pokestop while checking catchable pokemons. It will be useful when there are some pokestops with lure near each other.
But I don't know how long is the radius for catching pokemons.


jabbink:
Makes sense now why my bot fails to pick up heaps of pokemon while idling within 6 lured stops (distance 70m at most, as defined by the DOWNLOAD_SETTINGS): This API limits what it exposes even though I know there have to be more pokemon.

Will merge if you increase that range check to see if that pokemon is within 70m of you away.

mgorski-mg:
Ok, I'll do it in a minute.

AbandonedCart:
@jabbink Seriously. I was getting around 2 pokemonsters and I can stand right next to the stop and hold the lure like a sparkler. This is a huge improvement.

mgorski-mg:
Can we merge it? :)

jabbink:
haha just wanted to implement getting the settings myself. Nice job.

Will test now and merge (if it works)!

mgorski-mg:
Great! :)

library/src/main/java/com/pokegoapi/api/settings/FortSettings.java
```diff
@@ -1,10 +1,66 @@
...
+
+	@Getter
+	/**
+	 * NO IDEA
+	 *
+	 * @return distance in meters.
+	 */
+	private double farInteractionRangeMeters;
+
 	public void update(POGOProtos.Settings.FortSettingsOuterClass.FortSettings fortSettings) {
```
jabbink, line 58: JavaDoc missing

library/src/main/java/com/pokegoapi/api/map/fort/Pokestop.java
```diff
@@ -73,7 +73,19 @@ public boolean inRange() {
 		S2LatLng pokestop = S2LatLng.fromDegrees(getLatitude(), getLongitude());
 		S2LatLng player = S2LatLng.fromDegrees(api.getLatitude(), api.getLongitude());
 		double distance = pokestop.getEarthDistance(player);
-		return distance < 30;
+		return distance <= api.getSettings().getMapSettings().getPokemonVisibilityRange();
```
jabbink, line 76: Wrong setting

`api.getSettings().getFortSettings().getInteractionRangeInMeters()` (currently 40.0 which is the correct distance)

jabbink:
2 small issues, if you fix those Ill merge.

jabbink:
Anyways, merged.

PR #479, opened 2016-08-03
Importing for local build, Generalize add'l version numbers

AbandonedCart:
This includes instructions for importing the JAR when building locally and using gradle.

I am not sure it is necessary to provide such "thorough" documentation, but I am putting this here in case it is desired.


FabianTerhorst:
:+1:

PR #488, opened 2016-08-05
Add function to set the unknown6 field

jabbink:
**Changes made:**
- Add function to set the unknown6 field
  - temporary fix to make it accessible so other applications may set it
  - Don't use if you have no idea what this is for
  - Needs to be removed when it has been fully reversed, of course


FabianTerhorst:
Does it make sense to add a method that is getting removed in a few days anyway?

jabbink:
> in a few days 

I don't want to not bot for a few days. If it were a matter of minutes/hours I wouldn't bother. And as I will be using this code, and this is GPL 3, I thought, why not make it public. This introduces no breaking changes (null check on whether or not the value is set and if it's set, transmit it to the server).

PR #494, opened 2016-08-07
[WIP] add request signature

jabbink:
WIP

Todo
- [x] call encrypt
- [x] move signature stuff to different class


Paul776:
Nice, How will you calculate the sig?

jay8ee:
Are you going to try and generate it natively?

jabbink:
It does that right now. Here's a Java impl https://gist.github.com/jabbink/213e22c12db4273f0e7161eec014d327

jabbink:
But I have to go now, so feel free to change this PR/create a new one.

Shouldn't be too much work with the code in the changed files.

Paul776:
Possible bug: nuilder.getAuthTicket() always empty
it is never et anymore in ur PR

jabbink:
Fixed in the previous commit afaik

Paul776:
Just tried (with JNA imp) still get 0 to everything in the Map object
(python scripts still works)

Menes1337:
com/pokegoapi/util/Crypto.java:
`import javax.xml.bind.DatatypeConverter;`

seems not available in Android. I got an exception(java.lang.UnsatisfiedLinkError: Native library..) I think you just use it for debugging purpose. Thats why I just comment the lines out:

`//String hex = DatatypeConverter.printHexBinary(t);
//System.out.println(hex.toLowerCase());`

Paul776:
Something is wrong in getLocationHash1 you need to seed it with the ticket hash

vmarchaud:
@Paul776 yes already told him and got fixed in last commit.

library/src/main/java/com/pokegoapi/main/RequestHandler.java
```diff
@@ -313,4 +302,77 @@ public void run() {
...
+		}
+
+		// TODO: Call encrypt function on this
+		ByteString uk2 = sigBuilder.build().toByteString();
+		Unknown6OuterClass.Unknown6.newBuilder()
+				.setRequestType(6)
+				.setUnknown2(Unknown6OuterClass.Unknown6.Unknown2.newBuilder().setUnknown1(uk2));
+	}
+
+	private int getLocationHash1(byte[] authTicket, RequestEnvelope.Builder builder) {
```
Paul776: There is an other problem here, something is not right with the minus sign of the longitude

Paul776: I don't get the same result as if I use the original python script

jabbink: java uses signed int's. The python code unsigned.

Try to dump them both as hexadecimal (Integer.toHexString() in Java):

what is sent is exactly the same.

library/src/main/java/com/pokegoapi/main/RequestHandler.java
```diff
@@ -313,4 +304,94 @@ public void run() {
...
+		XXHashFactory factory = XXHashFactory.fastestInstance();
+		StreamingXXHash32 xx32 = factory.newStreamingHash32(0x1B845238);
+		xx32.update(authTicket, 0, authTicket.length);
+		xx32 = factory.newStreamingHash32(xx32.getValue());
+		byte[] bytes = new byte[8 * 3];
+
+		System.arraycopy(getBytes(api.getLatitude()), 0, bytes, 0, 8);
+		System.arraycopy(getBytes(api.getLongitude()), 0, bytes, 8, 8);
```
Paul776: Something is not right here with the minus sign of the location
if the python script xx32 hashed this -0.1277583 (with the 0x1B845238) it returns: 2878327542
here it return -1416639754

x32.hash(-0.1277583), 0, getBytes(-0.1277583).length, 0x1B845238)

(this is just a test to verify that the hashes are correct, as am still getting 0 pokemons back)

jabbink: As it is sent via protobuf I think that shouldn't matter, as it's the same memory representation

http://ideone.com/Y0DLIo

Paul776: I see.. something is not right somewhere then. Is it just me not getting anything back? Or is it working for you ?

jabbink: @vmarchaud has the solution, he says. This code is not working for anyone, as I've said in the comments (and this PR isn't merged, and it's called WIP).

iGio90: Im getting valid response with @jabbink impl btwn! WP

iGio90:
Im getting valid response! between, do you know if the other messages inside the signature (device infos etc) are also encr or not?

vmarchaud:
@iGio90 Currently nobody is sending them, they might force us to put them in the following days, we will see how to implement this.

iGio90:
@vmarchaud i saw your implementation today and i think its better to include these info directly in the impl. The UDID you used could be simply replaced with an Alphanum string of 40 lenght. All the other things you used are randomed good. It would be simple! do you know if they are encrypted or normal proto? i can make some test and PR in case!

vmarchaud:
@iGio90 removed it in the release, it was just to test last night. For the high level API we need to think a impl where user can put its own faked data.

Paul776:
http://imgur.com/a/Ez9yt

jabbink:
:tada:

jay8ee:
How can I actually get these changes and start testing? Can't seem to compile against the development branch with gradle.

jabbink:
Now it should be on jcenter

iGio90:
@vmarchaud could you please explain me why we should let the users insert data instead of sending fake ones? the purpose is just to prevent some sort of flags on missing device id. And also, 2 line serverside on device id will stop everything again! Still can't get why we couldn't send fake data

vmarchaud:
@iGio90 cause the problem is the same in your version, Niantic just has to ban ONE fake data to ban all our users, come on slack if you have more question.

iGio90:
@vmarchaud coming now! (edit: teamdomain?)

vmarchaud:
@iGio90 everything is at the bottom of the readme

library/src/main/java/com/pokegoapi/util/Signature.java
```diff
@@ -0,0 +1,115 @@
...
+				.setTimestamp(api.currentTimeMillis())
+				.setTimestampSinceStart(curTime - api.startTime);
+
+
+		for (RequestOuterClass.Request serverRequest : builder.getRequestsList()) {
+			byte[] request = serverRequest.toByteArray();
+			sigBuilder.addRequestHash(getRequestHash(authTicketBA, request));
+		}
+
+		// TODO: Call encrypt function on this
```
ScottPierce, line 51: @jabbink So what's to happen with this TODO?

PR #501, opened 2016-08-07
use safe implementation of XXHashFactory to make API workable on Android

mgorski-mg:


FabianTerhorst:
We could also use a check to determinate if the running platform is android, like okhttp, but maybe using the same for all platforms should be ok.

broach:
Tested on Android and is working perfectly 👍

AbandonedCart:
Checking for platform dependency is only a good option if the performance hit is significant. If this doesn't make a big difference and works, then it makes more sense.

broach:
Given the throttling on requests being so massive (in relative terms) I sincerely doubt a few microseconds is going to matter much when calculating the hash.

mgorski-mg:
I agree with you. What do you think @FabianTerhorst?

FabianTerhorst:
Is the speed worth the platform check?

mgorski-mg:
I don't think so, as @broach said few milliseconds is nothing comparing to api throttle.
But if you want we could check what is the time difference. But I can't do it right now.

jabbink:
I assume the `safeInstance` doesn't use the native implementations at all? Then maybe to decrease the size of this API only include the version without the JNI dep.

mgorski-mg:
@jabbink Do you think about changing the library to different which contains only safe java implementation? Or to shrink this one?

AbandonedCart:
Honestly, there is no real need for the entire lz4 library. xxhash is significantly smaller on its own.

vmarchaud:
@mgorski-zlatan Try implementing another lib or including only the safe instance (lz4 is under Apache2, we should be able to include it under ours by keeping the license header).

jabbink:
I then suggest stripping out the entire lz4 dep and include a Java impl of xxhash only.

I think it's weird it doesn't work on android though, as https://github.com/jpountz/lz4-java/blob/master/src/java/net/jpountz/xxhash/XXHashFactory.java#L128 wraps everything in try/catch, on fail loads the `fastestJavaInstance`, which in case of failure loads `safeInstance`.

FabianTerhorst:
@mgorski-zlatan you want to do what @jabbink is suggestion or should someone else doing it?

mgorski-mg:
Ok, I'll implement it.
@jabbink it's not working because android Unsafe class hasn't got getByte() method, so it can't be checked while initializing.

FabianTerhorst:
@mgorski-zlatan i would suggest you to create it as an different module.

jabbink:
(also make sure the picked impl actually works, many xxhash64 impls return garbage from what I've seen over the past days, xxhash32 seems to be implemented consistently over multiple platforms/languages)

kyr0:
Yeah, well I tried this and I'm getting:
java.lang.ClassNotFoundException: Didn't find class "net.jpountz.xxhash.XXHash32JavaSafe" on path

Any ideas? :)

My gradle includes dependency:

```
compile 'net.jpountz.lz4:lz4:1.3.0'
```

kyr0:
@mgorski-zlatan Just tell me if you want me to test/do anything; we are in the same boat :)

mgorski-mg:
If you want to help feel free. I didn't finish if yet, because I came from the work few minutes ago.

mgorski-mg:
@kyr0 If you have time today please feel free to change xxhash implementation, because I'll probably have no time today

kyr0:
Well, I figured out that the error disappeared as soon as I changed from:

```
buildTypes {
    debug {
        minifyEnabled true
        proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
    }
    release {
        minifyEnabled true
        proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
    }
}
```

To:

```
buildTypes {
    debug {
        minifyEnabled false
        proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
    }
    release {
        minifyEnabled false
        proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
    }
}
```

So disabling the minification process and thus proguard makes the jar's being available in the  resulting APK. 

**So seems like this isn't a library issue. It's a project-local Proguard rules issue (?)**

FabianTerhorst:
you have to add a proguard rule to keep net.jpountz.xxhash.XXHash32JavaSafe

broach:
Mine was already set to **minifyEnabled false** which is why I didn't run into further probs, sorry.

FabianTerhorst:
proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro' is useless when you disable it with false

FabianTerhorst:
btw this is definitely not a mistake from proguard, because proguard can´t know when an application needs an class and is only accessing this class with reflection.

kyr0:
> btw this is definitely not a mistake from proguard, because proguard can´t know when an application needs an class and is only accessing this class with reflection.

What I wanted to say: Local project configuration issue with proguard; for the time being I disabled it completely by setting minifyEnabled to false. (Later: Rules have to be added; so: Local proguard config issue) 

What was understood by you: proguard has an issue. 

Sorry, I never wanted to say that proguard has an issue.

FabianTerhorst:
`-keep public class net.jpountz.xxhash.XXHash32JavaSafe`  inside your proguard file should fit it.
Edit better: `-keep public class net.jpountz.xxhash.*`

cesco89:
@FabianTerhorst Thanks a lot!!!!
You are my hero! :D

cesco89:
Ok so, tried adding that flag to proguard and doesn't work.
On some devices keeps crashing (ZenFones and some other devices with android 5.x).
Also, i don't know if multidex is another crash cause with the updated library (4.0 up to date with latest commits)
i had to revert to an older version of the library to kill the 90% of the crashes

AbandonedCart:
```
-keepattributes *Annotation*
-keepattributes Signature
-keepattributes InnerClasses

-keep class org.apache.http.** { *; }
-keep class com.pokegoapi.** { *; }
-keep class POGOProtos.** { *; }
-keep class okhttp3.** { *; }
-keep class okio.** { *; }
-keep class com.google.protobuf.** { *; }
-keep class rx.** { *; }
-keep class net.jpountz.** { *; }

-dontwarn org.apache.http.**
-dontwarn com.pokegoapi.**
-dontwarn POGOProtos.**
-dontwarn okhttp3.**
-dontwarn okio.**
-dontwarn com.google.protobuf.**
-dontwarn rx.**
-dontwarn net.jpountz.**

-dontnote org.apache.http.**
-dontnote com.pokegoapi.**
-dontnote POGOProtos.**
-dontnote okhttp3.**
-dontnote okio.**
-dontnote com.google.protobuf.**
-dontnote rx.**
-dontnote net.jpountz.**

-keepattributes Exceptions,SourceFile,LineNumberTable
```

This is a sample ProGuard file from the application importing the library.
The last line is only if you want to have your native stacktrace deciphered.
(You can also use proguard tools to decipher a minified / obfuscated trace)

AbandonedCart:
Is it safe to assume we are all using this anyway?

PR #504, opened 2016-08-08
Add method for request Tutorial complete.

iGio90:
Some recent serverside changes now force the new accounts to accept tos before making requests. 

If any other can confirm, is working good for me!

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/PokemonGo.java
+++ b/library/src/main/java/com/pokegoapi/api/PokemonGo.java
@@ -15,8 +15,12 @@
 
 package com.pokegoapi.api;
 
+import POGOProtos.Enums.TutorialStateOuterClass;
 import POGOProtos.Networking.Envelopes.RequestEnvelopeOuterClass.RequestEnvelope.AuthInfo;
-import POGOProtos.Networking.Envelopes.Unknown6OuterClass;
+import POGOProtos.Networking.Requests.Messages.MarkTutorialCompleteMessageOuterClass;
+import POGOProtos.Networking.Requests.RequestTypeOuterClass.RequestType;
+import POGOProtos.Enums.TutorialStateOuterClass.TutorialState;
+
 import com.pokegoapi.api.inventory.Inventories;
 import com.pokegoapi.api.map.Map;
 import com.pokegoapi.api.player.PlayerProfile;
@@ -25,6 +29,7 @@ import com.pokegoapi.auth.CredentialProvider;
 import com.pokegoapi.exceptions.LoginFailedException;
 import com.pokegoapi.exceptions.RemoteServerException;
 import com.pokegoapi.main.RequestHandler;
+import com.pokegoapi.main.ServerRequest;
 import com.pokegoapi.util.SystemTimeImpl;
 import com.pokegoapi.util.Time;
 import lombok.Getter;
@@ -55,7 +60,6 @@ public class PokemonGo {
 	private CredentialProvider credentialProvider;
 	private Settings settings;
 	private Map map;
-	private List<Unknown6OuterClass.Unknown6> unknown6s = new ArrayList<>();
 
 	/**
 	 * Instantiates a new Pokemon go.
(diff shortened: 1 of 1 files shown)


broach:
Is this just fire-and-forget? When does it need to be called? What happens if the account has already accepted the TOS? 

Is there a server response sent when you try to use the API / log in telling you that you need to accept the TOS that this could be automatically sent as the reply?

AbandonedCart:
Depending if the server actually verifies, logs, or discards an overlapping attempt to accept will determine if this statement is relevant, but it appears be a fire and forget as part of the first login process (per account) by the user.
There should be a value provided by the server that it needs acceptance, so it would likely be possible to query the server if it required, but then you run into facing rate limits and such for something that is most likely going to be discarded by the endpoint if it is unnecessary.

broach:
I created a new PTC account and attempted to use this. No worky. 

The account authenticates and I get back the PtcCredentialProvider.  I instantiate a new PokemonGo and then call this new skipTos(). Subsequent API calls return empty; the account isn't past the TOS thing.

iGio90:
I just get awake and im gonna focus on it again! Im from phone now but i toke the idea from a python script linked in the issues. I think i replicated it right, but give it look please.

Anyway, I dont really know how they are doing it on server side because at the very first, i had valid response, then nothing as you said.

Edit: here is the script
https://github.com/keyphact/pgoapi/blob/691aff2543c39ba929e4aa0e325dc43d24cb24df/scripts/accept-tos.py

The request send tutorials_completed with value 0, maybe we have to use another value? ref:

https://github.com/AeonLucid/POGOProtos/blob/master/src/POGOProtos/Enums/TutorialState.proto

iGio90:
Anyway, I dont really know how they are doing it on server side because at the very first, i had valid response, then nothing as you said.

iGio90:
Update: The python script that i linked seems not working either. 
- It's not related to ToS. You can send valid request even without agree to tos.
- It will works once you log in with the official client, so you put the tutorial status to legal screen.

Repliceation: Create a new account either ptc/google. Run the call to set the tutorial into legal screen either with my impl or with the python script. Empty responses.
After the login with the official client it will fill the responses, even the one to put the tutorial into legal screen. Going for further investigation, is anyone having a look into?

Im going to MITM and check what happens at the first login

AbandonedCart:
It might be that by accepting the TOS, you are kicking into the tutorial and don't have a configuration to either eject or go through it. I dunno coming from an existing account, but it seems like before the bot would sidestep the entire process, but now it may be a callback (ie. if tos then avatar).

I could also be just rambling because I haven't looked at how python is actually doing it.

broach:
Yeah, I think to automate this is going to require looking at a MITM sequence for a new account and seeing what the game is doing.

iGio90:
It is! Im going ti retry once im back home! Today i tried both with xposed module and modified apk but didnt get Charles to work! If you are on it, just check what it goes out on the very first login of a fresh account! It happens before tos and before the call to set the status to login screen! (Almost sure)

iGio90:
Sorry i meant legal screen, cant edit comment with phone

iGio90:
Between login and ToS 

<-] Request for GetPlayer  { app_version: '0.31.0', '$unknownFields': [] }

[<-] Request for DownloadRemoteConfigVersion  { platform: 'ANDROID', app_version: 3100, '$unknownFields': [] }

[<-] Request for GetHatchedEggs  {}

[<-] Request for GetInventory  {}

[<-] Request for CheckAwardedBadges  {}

[<-] Request for DownloadSettings  {}

[<-] Request for DownloadItemTemplates  {}

[<-] Request for GetHatchedEggs  {}

[<-] Request for GetInventory  { last_timestamp_ms: '1470696953513', '$unknownFields': [] }

[<-] Request for CheckAwardedBadges  {}

[<-] Request for DownloadSettings  { hash: '6e9c84cbb761ec75078bb8f787559580b23a302b',
  '$unknownFields': [] }

iGio90:
@broach well.. i think Im done. It want this --> GetPlayer. Verified on 2 fresh accounts, it reply to the move to legal screen.

jay8ee:
Are you still working on this? Be nice to get this into dev

iGio90:
Of course! I finished it but im unable to remove this PR. Check my other PR (WIP tutorial state)

jay8ee:
Ah, I missed #522. Looks good :)

iGio90:
@jay8ee im waiting for some additional reports. Actually i tested it with more then 5 accts, but please, wanna hear something from others if it gives any kind of troubles

PR #508, opened 2016-08-08
Removed static mixed with non-static

Jazed:
Static modifier was conflicting, using this. with static modifier as well as rendering a function unusable.


FabianTerhorst:
👍  @jabbink merge please.

PR #510, opened 2016-08-08
make device infos settable

FabianTerhorst:


AbandonedCart:
A method to set device info is necessary, but this is a nightmare to implement.

```
DeviceInfos myInfo = new DeviceInfos() {
    @Override
    String getAndroidBoardName() {
        return myAndroidBoardName;
    }
    @Override
    String getAndroidBootloader() {
        return myAndroidBootloader;
    }

…
    @Override
    String getFirmwareType() {
        return myFirmwareType;
    }
    @Override
    String getFirmwareFingerprint() {
        return myFirmwareFingerprint;
    }
};
```

FabianTerhorst:
We need an interface for that and we don´t want duplicate ways for implementations.

FabianTerhorst:
We decided that, because we want to provide an interface for the platform.

FabianTerhorst:
updated the pr.

FabianTerhorst:
now its possible to use 

``` java
DeviceInfo deviceInfo = new DeviceInfo(); 
deviceInfo.set...("");
go.setDeviceInfo(deviceInfo)
```

library/src/main/java/com/pokegoapi/api/PokemonGo.java
```diff
@@ -192,4 +198,11 @@ public Map getMap() {
 		}
 		return map;
 	}
+
+	public SignatureOuterClass.Signature.DeviceInfo getDeviceInfo() {
```
jabbink, line 207: JavaDoc

PR #554, opened 2016-08-11
improve PokemonGo instance naming

FabianTerhorst:


AbandonedCart:
That fits with the more recent changes, but instance should probably still be instance to avoid confusion.

FabianTerhorst:
On more then 90% of the files its named api so i thought this was a decision from someone.

FabianTerhorst:
@Grover-c13 rename them all to instance or to api ?

AbandonedCart:
Yeah, I just realized where that was located. I thought it was some sort of cached instance. Nevermind

FabianTerhorst:
Its just the object reference.

AbandonedCart:
Yeah. I was looking at the egg hatching and the excerpt looked like it was a different file. I was thinking a cached instance, like when an app makes a static instance that it keeps private. It was just another per-class reference with a different name, though.

Issue #558, opened 2016-08-11
NoSuchItemException because Inventory is not updated forcely on startup


olegshulyakov:
**Description:**
Got **com.pokegoapi.exceptions.NoSuchItemException** when trying to catch pokemon.

**Reason:** 
_lastInventoryUpdate_ is 0.
This is because Inventory is initialed lazy. There is no force update in construction **_Inventories:74_**

**Stacktrace (If it's a crash):**
com.pokegoapi.exceptions.NoSuchItemException
    at com.pokegoapi.api.map.pokemon.CatchablePokemon.getItemBall(CatchablePokemon.java:285) ~[PokeGOAPI-library-0.4.0.jar:?]
    at com.pokegoapi.api.map.pokemon.CatchablePokemon.catchPokemon(CatchablePokemon.java:604) ~[PokeGOAPI-library-0.4.0.jar:?]

**Version:**
0.4.0

PR #559, opened 2016-08-11
Issue 558

olegshulyakov:
**Fixed issue:** [https://github.com/Grover-c13/PokeGOAPI-Java/issues/558]

**Changes made:**
- removed lazy initialization of inventory
- migrated to concurrent collections in inventory package

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/PokemonGo.java
+++ b/library/src/main/java/com/pokegoapi/api/PokemonGo.java
@@ -51,8 +51,9 @@ public class PokemonGo {
 	@Getter
 	RequestHandler requestHandler;
 	@Getter
-	private PlayerProfile playerProfile;
-	private Inventories inventories;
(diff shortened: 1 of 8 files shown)


FabianTerhorst:
We had at the start a http request for a force load, but i think we decided to drop this, because of the http request on object creation.

olegshulyakov:
I thought about using cache for inventory and update it like map objects

AbandonedCart:
In my tests, this actually causes a warning (and login failure) where there wasn't one before

FabianTerhorst:
I can login with this pull request, i can also catch a pokemon. Most of your changes are bad for the performance. Fro example: CopyOnWriteArraySet is slower then HashSet or ConcurrentMap is slower then HashMap.

FabianTerhorst:
The most concurrent exceptions you can prevent by creating a new array or hashmap from the returned value instead of using the reference.

PR #561, opened 2016-08-11
Add Native Walking API

mackenzieg:
Just wanted to add some native walking to the APi, should add a little bit of randomness to to the walking as to not trigger the server detecting our perfect locations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,7 +15,7 @@ serve as a standard template for any Bapcraft Java projects.
 .mtj.tmp/
 
 # Package Files #
-*.jar
+.jar
 *.war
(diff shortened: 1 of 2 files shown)


FabianTerhorst:
You can use rxjava timer functions to make it fit better into the project.

AbandonedCart:
Isn't this more for the individual client?
I know in my case, with a mock location client included, this is overhead.

It would be great for the sample project, though.

FabianTerhorst:
you can create a sample with a walking logic.

Grover-c13:
this is outside the scope of the API sorry.

PR #566, opened 2016-08-12
Add throws when account appears to be banned

Grover-c13:
Add throws when account appears to be banned


pvanassen:
Why not a BannedException or something of that kind so api users can catch it and handle it properly?

Paul776:
+1...

Grover-c13:
Because status code 3 is not exclusive for bans, for example sending a request with invalid fields will result in a 3. Ideally this wouldnt happen in the API but still can, and until that time we will keep it like this. Its also why the message says you may be banned

FabianTerhorst:
Maybe we should just return the status codes in the exception messenges so that the applications can catch them and getMessage() == 3 //Maybe ban.

AbandonedCart:
Even with a banned account, you can get a contents of buffer null exception before it has the chance to return this one. My opinion is to leave it to the user, since it sounds too unreliable.

Paul776:
Leave to the use but how?  How do you check the account is banned?  You need to retrieve the cause and then the message and then compared it to see if it contains the word ban.  This totally unacceptable...

PR #587, opened 2016-08-15
Tutorial state check

iGio90:
This will allows new accounts to receive valid responses from PoGo. With the latest changes on PoGo apis, new accounts now require to set the tutorial state to LEGAL_SCREEN.
In addition to this, new accounts require to send the GET_PLAYER requests before mark the tutorial state.
The code check for the presence of the required tutorial state and, if not present, it will set it up.
Feel free to merge or suggest any additional modification required.


iGio90:
@benjy3gg here you are 💃

benjy3gg:
so, what do i have to call after login now, to get it working right?

PR #588, opened 2016-08-15
verify stardust requirement in canPowerUp

AbandonedCart:
This verifies the user can actually "afford" to perform the power-up

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/pokemon/Pokemon.java
+++ b/library/src/main/java/com/pokegoapi/api/pokemon/Pokemon.java
@@ -20,6 +20,7 @@ import com.google.protobuf.InvalidProtocolBufferException;
 import com.pokegoapi.api.PokemonGo;
 import com.pokegoapi.api.inventory.Item;
 import com.pokegoapi.api.map.pokemon.EvolutionResult;
+import com.pokegoapi.api.player.PlayerProfile;
 import com.pokegoapi.exceptions.AsyncRemoteServerException;
 import com.pokegoapi.exceptions.LoginFailedException;
 import com.pokegoapi.exceptions.RemoteServerException;
@@ -172,7 +173,8 @@ public class Pokemon extends PokemonDetails {
 	 * @throws RemoteServerException the remote server exception
 	 */
 	public boolean canPowerUp() throws LoginFailedException, RemoteServerException {
-		return getCandy() >= getCandyCostsForPowerup();
+		return getCandy() >= getCandyCostsForPowerup() && api.getPlayerProfile()
+			.getCurrency(PlayerProfile.Currency.STARDUST) >= getStardustCostsForPowerup();
 	}
 
 	/**


FabianTerhorst:
rebase please from the latest commit to fix build error.

LoungeKatt:
Relax, the commit came 20 minutes after the PR.

library/src/main/java/com/pokegoapi/api/pokemon/Pokemon.java
```diff
@@ -170,9 +172,12 @@ public Result transferPokemon() throws LoginFailedException, RemoteServerExcepti
 	 * @return the boolean
 	 * @throws LoginFailedException  the login failed exception
 	 * @throws RemoteServerException the remote server exception
+	 * @throws InvalidCurrencyException the currency has exception
 	 */
-	public boolean canPowerUp() throws LoginFailedException, RemoteServerException {
-		return getCandy() >= getCandyCostsForPowerup();
+	public boolean canPowerUp() throws LoginFailedException, RemoteServerException, InvalidCurrencyException {
+		return getCandy() >= getCandyCostsForPowerup()
```
FabianTerhorst: get currency don´t throw a InvalidCurrencyException

AbandonedCart: Please read the entire change. `api.getPlayerProfile().getCurrency(PlayerProfile.Currency.STARDUST)` does.

FabianTerhorst: http://imgur.com/a/XWQFb

AbandonedCart: Take out the exception and compile it. You'll get an error. Why do you think both PRs handled that exception in one way or another? If you feel there shouldn't be an exception, then fix the original class to stop passing it to the reference. That isn't within the scope of this PR.

FabianTerhorst: There is no exception in the original class.

FabianTerhorst: https://github.com/Grover-c13/PokeGOAPI-Java/blob/Development/library/src/main/java/com/pokegoapi/api/player/PlayerProfile.java#L211

mgorski-mg: But it was an exception. I already fix it in last PR. 
@LoungeKatt pull last changes from Development and eveything should work without exceptions.

AbandonedCart: @FabianTerhorst The exception was removed 2 minutes before your comment. Luckily, @mgorski-zlatan mentioned that a recent change has modified the PR without presenting an error. I am not sure why there had to be this elaborate claim a commit from a day ago did not account for a change 45 minutes ago without simply referencing the change.

PR #597, opened 2016-08-16
Added/Modified some methods in Pokemon.java

gionata-bisciari:
Changes made:
+Added canEvolve():boolean
*Modified canPowerUp():boolean (now it checks if the player has enough star dust to powerup)

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/pokemon/Pokemon.java
+++ b/library/src/main/java/com/pokegoapi/api/pokemon/Pokemon.java
@@ -20,7 +20,9 @@ import com.google.protobuf.InvalidProtocolBufferException;
 import com.pokegoapi.api.PokemonGo;
 import com.pokegoapi.api.inventory.Item;
 import com.pokegoapi.api.map.pokemon.EvolutionResult;
+import com.pokegoapi.api.player.PlayerProfile.Currency;
 import com.pokegoapi.exceptions.AsyncRemoteServerException;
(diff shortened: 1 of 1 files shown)


AbandonedCart:
This may be a partial overlap of https://github.com/Grover-c13/PokeGOAPI-Java/pull/588

gionata-bisciari:
It's based on the latest commit, and I think that canPowerUp() shouldn't throw an InvalidCurrencyException  because it gets the currency name directly from Currency class, maybe I'm wrong but I think that

AbandonedCart:
It's typically bad practice to ignore an exception and it passes along any issue with the currency from the server (not just the variable definition), but both PRs are based on the latest commit.

FabianTerhorst:
The method should throw the exceptions, because we don´t ignore exceptions.

library/src/main/java/com/pokegoapi/api/pokemon/Pokemon.java
```diff
@@ -172,7 +174,30 @@ public Result transferPokemon() throws LoginFailedException, RemoteServerExcepti
 	 * @throws RemoteServerException the remote server exception
 	 */
 	public boolean canPowerUp() throws LoginFailedException, RemoteServerException {
-		return getCandy() >= getCandyCostsForPowerup();
+		boolean can = false;
```
FabianTerhorst, line 177: remove the changes to the canPowerUp method and this could be merged.

gionata-bisciari:
So, I'll close it.

PR #606, opened 2016-08-16
Add an interface to PokemonIdValue from CatchablePokemon

AbandonedCart:
This allows referencing the individual class getPokemonIdValue directly from the CatchablePokemon object

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/map/pokemon/CatchablePokemon.java
+++ b/library/src/main/java/com/pokegoapi/api/map/pokemon/CatchablePokemon.java
@@ -84,6 +84,8 @@ public class CatchablePokemon implements MapPoint {
 	@Getter
 	private final PokemonId pokemonId;
 	@Getter
+	private final int pokemonIdValue;
+	@Getter
 	private final long expirationTimestampMs;
 	@Getter
 	private final double latitude;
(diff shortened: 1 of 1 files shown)


FabianTerhorst:
Is the pokemon id value the same as CatchablePokemon cp; cp.getPokemonId().getNumber() ?

FabianTerhorst:
the proto java file is wrapping the id from getPokemonIdValue to create the enum object for getPokemonId()

AbandonedCart:
MapPokemon pokemon.getPokemonIdValue()
WildPokemon:  pokemon.getPokemonData().getPokemonIdValue()
FortLureInfo: lureInfo.getActivePokemonIdValue()

Whether or not it is provided through an alternate means, this makes it uniform with the rest of the API, which offer both values individually

FabianTerhorst:
We have never wrapped the value int from the proto file to the wrappers before.

AbandonedCart:
Are you trying to say the type is wrong or is there some other issue other? If you meant the type was wrong, that can easily be changed. It sounds like you were against the entire idea, though, which honestly makes no sense.

FabianTerhorst:
I mean you are the one that told us that this project has to much methods and now you want to add a duplicate way to access a value.

AbandonedCart:
So there isn't an issue with the PR? Because the type is correct.

Int

FabianTerhorst:
I never told that the type is incorrect. I just said that this is a duplicate way to access the value.

AbandonedCart:
`We have never wrapped the value int from the proto file to the wrappers before` was a little confusing then, since all of them are int values.

The merit of `getPokemonIdValue` is that when used in an application that has it's own collection of references, for example images, you avoid the overhead of getting the wrapped object and then unwrapping it again to make an already expensive call to get the image for that pokedex value.

If the choice is between one getter in an API that already forces MultiDex or the overhead of having to make up to 151 calls to unwrap a wrapped value only to avoid that one method, it's a no-brainer.

PR #607, opened 2016-08-16
Added new useful methods to Pokemon.java

gionata-bisciari:
+getStaminaInPercentage():int
+getCPInPercentageActualPlayerLevel():int
+getCPInPercentageMaxPlayerLevel:int
+getIvInPercentage():double

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/pokemon/Pokemon.java
+++ b/library/src/main/java/com/pokegoapi/api/pokemon/Pokemon.java
@@ -418,7 +418,7 @@ public class Pokemon extends PokemonDetails {
 	 * @throws LoginFailedException if threw from {@link #getMaxCpForPlayer()}
 	 * @throws RemoteServerException if threw from {@link #getMaxCpForPlayer()}
 	 */
-	public int getCPInPercentageActualPlayerLevel() throws NoSuchItemException, LoginFailedException, RemoteServerException{
+	public int getCPInPercentageActualPlayerLevel() throws NoSuchItemException, LoginFailedException, RemoteServerException {
 		return ((getCp()*100)/getMaxCpForPlayer());
 	}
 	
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/pokegoapi/api/pokemon/Pokemon.java
```diff
@@ -403,4 +404,36 @@ public boolean isFainted() {
...
+	 * @throws NoSuchItemException if threw from {@link #getMaxCp()}
+	 */
+	public int getCPInPercentageMaxPlayerLevel() throws NoSuchItemException{
+		return ((getCp()*100)/getMaxCp());
+	}
+	
+	/**
+	 * @return IV in percentage
+	 */
+	public double getIvInPercentage(){
```
FabianTerhorst: Add Space between ) and {

AbandonedCart:
That's a nice way to interpret those values, especially when comparing them.
Too bad at anything below 35%, it can't just return `NAME fails as a Pokemon`

Out of curiosity, why does the final one return a double?
As far as I know, Math.floor will already return an int.

gionata-bisciari:
ivs
it returns a double with two decimal places

AbandonedCart:
My mistake. Missed the second outer parenthesis for the / 100

FabianTerhorst:
Please fix the style warnings.

gionata-bisciari:
what style warnings?

FabianTerhorst:
When you run the gradle task on your project.

FabianTerhorst:
I still don´t know why travis doesn´t print them.

gionata-bisciari:
I haven't seen it infact, I'll check

FabianTerhorst:
👍

Grover-c13:
yeah i really wish we could get travis to fail on checkstyle :(

gionata-bisciari:
Is now everything ok?

PR #608, opened 2016-08-16
One size fits all catch options

AbandonedCart:
Usage:

```
CatchOptions options = new CatchOptions(go);
options.maxRazzberries(5);
options.useBestBall(true);
options.noMasterBall(true);

cp.catchPokemon(options);
```

The options are all optional and each one overrides any with a similar
functionality based on the most relevant option, for example
usePokeBall can be set as a minimum by using it with useBestBall, a
maximum by using it alone, or exclusive by using with noFallback


Grover-c13:
Why not return the CatchOptions object on each method for builder style?

AbandonedCart:
I was aiming for that, but I can't find an example that isn't referencing the protos to fully understand how that should actually look. Do you know of one off-hand?

Nevermind, I just searched the builder pattern instead. Hopefully it comes out right, lol.

FabianTerhorst:
That's chaining. A builder is a inner class that is using chaining. The build method is creating the object with the constructor.

AbandonedCart:
@FabianTerhorst http://www.javaworld.com/article/2074938/core-java/too-many-parameters-in-java-methods-part-3-builder-pattern.html

FabianTerhorst:
Yes and your isn't a builder, because you use the same class for chaining and for creating.

LoungeKatt:
@FabianTerhorst It is just a label on the commit. If you have any real issues, please post those here. If you intend to nitpick a commit message, there is a slack channel for that.

LoungeKatt:
TODO: 
convert asynchronous BestBallToUse
add parameters for throwing balls
merge
handle backlash

AbandonedCart:
Time to cause some havoc...

FabianTerhorst:
Better squash the commits next time.

AbandonedCart:
Coming from you? Seriously? Each commit is an individual item.

FabianTerhorst:
You can still squash them to don´t destroy the history.

AbandonedCart:
Yes, or they can be left as individual items with a different purpose.

FabianTerhorst:
We squashed all commits the last weeks you are the only one that doesn´t squashed the commits.

AbandonedCart:
You are comparing "added a single function" to "rewrote the entire structure"

FabianTerhorst:
Its would be still better to see one feature/commit.

LoungeKatt:
Ok, done.

LoungeKatt:
1. Constructor
2. Rare use cases
3. Single use items
4. Javadocs and variable names
5. Replaced method docs
6. Builder design + Move to settings could have been merged
7. Async
8. Link all javadocs
9. Replace BestBallToUse + Async version could have been merged
10. Hit parameters
11. README

So of 13 commits to complete the entire master feature of rewriting the CatchPokemon structure, there were two places where they could be considered a single feature and squashed, cutting the PR down by two commits. Are 2 commits really worth the amount of time and effort you'll waste?

PR #623, opened 2016-08-18
Development

gionata-bisciari:
*Fixed Pokemon.canEvolve():boolean, now checks if the pokemon reached the maximum evolution form;
+Added Pokemon.canPowerUp(considerMaxCPLimitForPlayerLevel:boolean):boolean so now passing true it checks if the pokemon already reached its current maximum cp for player level


AbandonedCart:
It would probably be easier to rebase this instead of adding the existing commits and then having to patch them to fit the current state of the repository.

AbandonedCart:
You can also clone https://github.com/AbandonedCart/PokeGOAPI-Java/tree/evolution and use that to submit the pull request if that will make it easier.

gionata-bisciari:
I didn't know, sorry, and thanks

AbandonedCart:
No need to apologize. It's just easier to look at that way and avoids a few extra commits. 
It wouldn't cause any real issues, though.

PR #634, opened 2016-08-19
Fix razz berries usage

gsisso:
**Changes made:**
in the old code if useRazzBerries was set to false then it would either return maxRazzBerries if maxRazzBerries > 0 or -1 (unlimited) if set to 0 or less.
atm theres no way to use CatchOptions object without using razz berries.

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/settings/CatchOptions.java
+++ b/library/src/main/java/com/pokegoapi/api/settings/CatchOptions.java
@@ -198,8 +198,8 @@ public class CatchOptions {
 	 */
 	public int getRazzberries() {
 		int razzberries = maxRazzBerries;
-		if (useRazzBerries && maxRazzBerries <= 1) {
-			razzberries = 1;
+		if(!useRazzBerries){
+			return 0;
 		} else if (maxRazzBerries <= 0) {
 			razzberries = -1;
 		} else {


FabianTerhorst:
fix checkStyle please

AbandonedCart:
It appears the original was incorrect, as the default for catchPokemon without razzberries was to set it to -1. I have a feeling this is also incorrect, but I need to verify.

gsisso:
@FabianTerhorst Sorry I used the github raw edit ,would appreciate if you could just close this pr and make the change your self.

@LoungeKatt could you explain what do you mean ? I set it that if useRazzBerries  is false then it wouldn't use any razz berries and if set to true and maxRazzBerries is set to 0 or less it would return unlimited otherwise it would return maxRazzBerries

AbandonedCart:
The correction you made will cause it to fail elsewhere. I'll submit a PR that will explain a bit better.

The idea is that useRazzberries true will supersede maxRazzberries, but not useRazzberries false, since that may not accompany setting useRazzberries at all. By your method, you have to set both.

AbandonedCart:
Hopefully that makes sense. With your change, the default would still be unlimited (rather than none) and you would create the scenario where setting a max of 0 without setting useRazzberries true would also set unlimited, which kills off the possibility to achieve the same outcome setting max as 0.

AbandonedCart:
Thank you for pointing out the issue. That also explains a long history of why there was no actual default option to use "nothing" without calling a "WithRazzberries" variant that set a value of 0.

PR #636, opened 2016-08-19
Fix maxRazzberries for intended functionality

AbandonedCart:
Performs the intended fix of https://github.com/Grover-c13/PokeGOAPI-Java/pull/634 without changing the functionality

This should allow setting true when no max exists to use 1, but allow
setting a max of more than 1 to supersede leaving use as false

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/settings/CatchOptions.java
+++ b/library/src/main/java/com/pokegoapi/api/settings/CatchOptions.java
@@ -66,7 +66,7 @@ public class CatchOptions {
 	public CatchOptions(PokemonGo api) {
 		this.api = api;
 		this.useRazzBerries = false;
-		this.maxRazzBerries = -1;
+		this.maxRazzBerries = 0;
 		this.useBestPokeball = false;
 		this.skipMasterBall = false;
 		this.pokeBall = POKEBALL;
@@ -197,15 +197,7 @@ public class CatchOptions {
 	 * @return the number to use
 	 */
 	public int getRazzberries() {
-		int razzberries = maxRazzBerries;
-		if (useRazzBerries && maxRazzBerries <= 1) {
-			razzberries = 1;
-		} else if (maxRazzBerries <= 0) {
-			razzberries = -1;
-		} else {
-			razzberries = maxRazzBerries;
-		}
-		return razzberries;
+		return useRazzBerries && maxRazzBerries == 0 ? 1 : maxRazzBerries;
 	}
 	
 	/**


gsisso:
still doesn't fix the issue that its impossible to set razz berries to 0  - when I dont want it to use any razz berries.

LoungeKatt:
How so? You can set max to 0 or use to false. If you set neither, you use none because those are the defaults. I think that is the part you may be missing. The constructor sets them initially and you override them with your own options.

AbandonedCart:
`if (useRazzBerries && maxRazzBerries == 0)` means if useRazzBerries is false, you default to `maxRazzBerries` which is 0. If you set `maxRazzBerries` to 0, it leaves `useRazzBerries` as false, so you default to `maxRazzBerries` which is 0. If you set neither, you default to `maxRazzBerries` which is 0. No matter how you go about it, they will all give you 0 as the number of razzberries actually used. 

Make sense?

library/src/main/java/com/pokegoapi/api/settings/CatchOptions.java
```diff
@@ -198,10 +198,8 @@ public Pokeball getItemBall(double encounterProbability) throws LoginFailedExcep
 	 */
 	public int getRazzberries() {
 		int razzberries = maxRazzBerries;
```
FabianTerhorst: int razzberries = maxRazzBerries; is a redundant call.

AbandonedCart:
In code terms:

```
CatchOptions options = new CatchOptions(go);
cp.catchPokemon(options);
```

```
CatchOptions options = new CatchOptions(go).useRazzberries(false);
cp.catchPokemon(options);
```

```
CatchOptions options = new CatchOptions(go).maxRazzberries(0);
cp.catchPokemon(options);
```

```
CatchOptions options = new CatchOptions(go).useRazzberries(false).maxRazzberries(0);
cp.catchPokemon(options);
```

will not use any razzberries.

library/src/main/java/com/pokegoapi/api/settings/CatchOptions.java
AbandonedCart: No, it's setting the local variable as the class variable. The else would be redundant.

FabianTerhorst: int razzberries = 42; has the same outcome.

FabianTerhorst: you are setting the value on every if statement

FabianTerhorst: you can simplify this to 

``` java
return useRazzBerries && maxRazzBerries == 0
                ? 1
                : maxRazzBerries;
```

FabianTerhorst: this would reduce a cleanup for the gc.

LoungeKatt: Is that your final suggestion? It is rather tedious to attempt to adjust as the comments flood in.

FabianTerhorst:
wouldn´t be 

```
public int getRazzberries() {
        return useRazzBerries && maxRazzBerries == 0 ? 1 : maxRazzBerries;
}
```

not much more readable?

FabianTerhorst:
👍

AbandonedCart:
Patience is a virtue...

AbandonedCart:
@gsplash Does this allow for what you were trying to accomplish?

gsisso:
yes seem like it should work fine now.

AbandonedCart:
@gsplash Again, **thank you** for pointing that out. I had doubts about the original defaults and without your PR, it would have carried over. The whole point of this overhaul was to start cleaning out issues like that by using something a little more organized / universal.

PR #642, opened 2016-08-19
Adaptive ball selection option

AbandonedCart:
This will use the user parameters to attempt selecting a ball, but clear the restrictions and retry to prevent throwing an inventory exception.

This resolves the issue where parameters are set, but the user would prefer overriding the parameters instead of throwing the inventory error.

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/settings/AsyncCatchOptions.java
+++ b/library/src/main/java/com/pokegoapi/api/settings/AsyncCatchOptions.java
@@ -22,6 +22,7 @@ import com.pokegoapi.exceptions.LoginFailedException;
 import com.pokegoapi.exceptions.NoSuchItemException;
 import com.pokegoapi.exceptions.RemoteServerException;
 
+import static POGOProtos.Inventory.Item.ItemIdOuterClass.ItemId;
 import static POGOProtos.Inventory.Item.ItemIdOuterClass.ItemId.ITEM_GREAT_BALL;
 import static POGOProtos.Inventory.Item.ItemIdOuterClass.ItemId.ITEM_MASTER_BALL;
 import static POGOProtos.Inventory.Item.ItemIdOuterClass.ItemId.ITEM_POKE_BALL;
@@ -30,6 +31,8 @@ import static com.pokegoapi.api.inventory.Pokeball.GREATBALL;
 import static com.pokegoapi.api.inventory.Pokeball.MASTERBALL;
 import static com.pokegoapi.api.inventory.Pokeball.POKEBALL;
 import static com.pokegoapi.api.inventory.Pokeball.ULTRABALL;
+
+import java.util.Arrays;
 
 import lombok.Getter;
 import lombok.ToString;
@@ -48,6 +51,7 @@ public class AsyncCatchOptions {
 	private int useRazzBerry;
 	private Pokeball pokeBall;
 	private boolean strictBallType;
+	private boolean smartSelect;
 	private double probability;
 	@Getter
 	private double normalizedHitPosition;
@@ -68,6 +72,7 @@ public class AsyncCatchOptions {
 		this.skipMasterBall = false;
 		this.pokeBall = POKEBALL;
 		this.strictBallType = false;
+		this.smartSelect = false;
 		this.probability = 0;
 		this.normalizedHitPosition = 1.0;
 		this.normalizedReticleSize = 1.95 + Math.random() * 0.05;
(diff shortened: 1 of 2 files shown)


gsisso:
instead of duplicating 

```
            if (smartSelect) {
                useBestPokeball = false;
                skipMasterBall = false;
                smartSelect = false;
                return getItemBall();
            }
            throw new NoSuchItemException();
```

put it after the if/else statement (at the end of the method)

gsisso:
also you have a typo at -

> Set whether or not **_ot_** use adaptive ball selection

AbandonedCart:
@gsplash Each section has a throw at the end. Putting it at the end would result in it never being reached. Removing the throw would invalidate the purpose of exclusive ball selection (just in case that was the next statement)

gsisso:
@LoungeKatt thats why I said put the throw at the end.

instead of -

```
    public Pokeball getItemBall() throws LoginFailedException,
                        RemoteServerException, NoSuchItemException {
        ItemBag bag = api.getInventories().getItemBag();
        if (strictBallType) {
            if (bag.getItem(pokeBall.getBallType()).getCount() > 0) {
                return pokeBall;
            } else if (useBestPokeball) {
                if (!skipMasterBall) {
                    if (bag.getItem(ITEM_MASTER_BALL).getCount() > 0) {
                        return MASTERBALL;
                    }
                } else if (bag.getItem(ITEM_ULTRA_BALL).getCount() > 0) {
                    return ULTRABALL;
                }
            } else if (bag.getItem(ITEM_POKE_BALL).getCount() > 0) {
                return POKEBALL;
            }
            if (smartSelect) {
                strictBallType = false;
                useBestPokeball = false;
                skipMasterBall = false;
                smartSelect = false;
                return getItemBall();
            }
            throw new NoSuchItemException();
        } else {
            int index = 3;
            if (pokeBall.getBallType() == ITEM_MASTER_BALL) index = 3;
            if (pokeBall.getBallType() == ITEM_ULTRA_BALL) index = 2;
            if (pokeBall.getBallType() == ITEM_GREAT_BALL) index = 1;
            if (pokeBall.getBallType() == ITEM_POKE_BALL) index = 0;

            if (useBestPokeball) {
                if (!skipMasterBall && index <= 3 && bag.getItem(ITEM_MASTER_BALL).getCount() > 0) {
                    return MASTERBALL;
                } else if (index <= 2 && bag.getItem(ITEM_ULTRA_BALL).getCount() > 0) {
                    return ULTRABALL;
                } else if (index <= 1 && bag.getItem(ITEM_GREAT_BALL).getCount() > 0) {
                    return GREATBALL;
                } else if (bag.getItem(ITEM_POKE_BALL).getCount() > 0) {
                    return POKEBALL;
                }
            } else {
                if (index >= 0 && bag.getItem(ITEM_POKE_BALL).getCount() > 0) {
                    return POKEBALL;
                } else if (index >= 1 && bag.getItem(ITEM_GREAT_BALL).getCount() > 0) {
                    return GREATBALL;
                } else if (index >= 2 && bag.getItem(ITEM_ULTRA_BALL).getCount() > 0) {
                    return ULTRABALL;
                } else if (!skipMasterBall && index <= 0 && bag.getItem(ITEM_MASTER_BALL).getCount() > 0) {
                    return MASTERBALL;
                }
            }
            if (smartSelect) {
                useBestPokeball = false;
                skipMasterBall = false;
                smartSelect = false;
                return getItemBall();
            }
            throw new NoSuchItemException();
        }
    }
```

do -

```
    public Pokeball getItemBall() throws LoginFailedException,
            RemoteServerException, NoSuchItemException {
        ItemBag bag = api.getInventories().getItemBag();
        if (strictBallType) {
            if (bag.getItem(pokeBall.getBallType()).getCount() > 0) {
                return pokeBall;
            } else if (useBestPokeball) {
                if (!skipMasterBall) {
                    if (bag.getItem(ITEM_MASTER_BALL).getCount() > 0) {
                        return MASTERBALL;
                    }
                } else if (bag.getItem(ITEM_ULTRA_BALL).getCount() > 0) {
                    return ULTRABALL;
                }
            } else if (bag.getItem(ITEM_POKE_BALL).getCount() > 0) {
                return POKEBALL;
            }
        } else {
            int index = 3;
            if (pokeBall.getBallType() == ITEM_MASTER_BALL) index = 3;
            if (pokeBall.getBallType() == ITEM_ULTRA_BALL) index = 2;
            if (pokeBall.getBallType() == ITEM_GREAT_BALL) index = 1;
            if (pokeBall.getBallType() == ITEM_POKE_BALL) index = 0;

            if (useBestPokeball) {
                if (!skipMasterBall && index <= 3 && bag.getItem(ITEM_MASTER_BALL).getCount() > 0) {
                    return MASTERBALL;
                } else if (index <= 2 && bag.getItem(ITEM_ULTRA_BALL).getCount() > 0) {
                    return ULTRABALL;
                } else if (index <= 1 && bag.getItem(ITEM_GREAT_BALL).getCount() > 0) {
                    return GREATBALL;
                } else if (bag.getItem(ITEM_POKE_BALL).getCount() > 0) {
                    return POKEBALL;
                }
            } else {
                if (index >= 0 && bag.getItem(ITEM_POKE_BALL).getCount() > 0) {
                    return POKEBALL;
                } else if (index >= 1 && bag.getItem(ITEM_GREAT_BALL).getCount() > 0) {
                    return GREATBALL;
                } else if (index >= 2 && bag.getItem(ITEM_ULTRA_BALL).getCount() > 0) {
                    return ULTRABALL;
                } else if (!skipMasterBall && index <= 0 && bag.getItem(ITEM_MASTER_BALL).getCount() > 0) {
                    return MASTERBALL;
                }
            }
        }
        if (smartSelect) {
            useBestPokeball = false;
            skipMasterBall = false;
            smartSelect = false;
            return getItemBall();
        }
        throw new NoSuchItemException();
    }
```

AbandonedCart:
@gsplash You said to move the smartSelect to the end. Adding in to also put the throw at the end makes more sense.

LoungeKatt:
That isn't done. Still fixing the reverse of the ball array

library/src/main/java/com/pokegoapi/api/settings/AsyncCatchOptions.java
```diff
@@ -151,23 +161,22 @@ public Pokeball getItemBall(double encounterProbability) throws LoginFailedExcep
...
 			}
 			throw new NoSuchItemException("No ball available for capture probability");
 		} else {
-			int index = 3;
-			if (pokeBall.getBallType() == ITEM_MASTER_BALL) index = 3;
-			if (pokeBall.getBallType() == ITEM_ULTRA_BALL) index = 2;
-			if (pokeBall.getBallType() == ITEM_GREAT_BALL) index = 1;
-			if (pokeBall.getBallType() == ITEM_POKE_BALL) index = 0;
+			ItemId[] pokeballs = new ItemId[] { ITEM_MASTER_BALL, ITEM_ULTRA_BALL,
+					ITEM_GREAT_BALL, ITEM_POKE_BALL };
```
FabianTerhorst: You could just create a ArrayList with the values.

AbandonedCart: Then you had to chain .add commands together. I thought you were all about readability?

FabianTerhorst: ArrayList<String> list = new ArrayList<String>() {{
    add("A");
    add("B");
    add("C");
}};

AbandonedCart: That is chaining. This saves lines.

FabianTerhorst: It reduces a value that has get cleaned up from the gc.

AbandonedCart: Not true. Both only set 1 variable. The other is discarded upon use.
Thanks for the suggestion, though.

FabianTerhorst: Is Array as list faster then creating a new ArrayList when you Benchmark it?

AbandonedCart: Not sure. You go test it and let me know.

gsisso:
why creating the array every method call ? this could be added as a constant.
also maybe instead of creating an array and then convert it to list maybe just initial it as a list in the first place. 

`private static final List<ItemId> pokeballs  = Arrays.asList(ITEM_MASTER_BALL, ITEM_ULTRA_BALL,ITEM_GREAT_BALL, ITEM_POKE_BALL);`

then you can do
`int index = pokeballs.indexOf(pokeBall.getBallType());`

AbandonedCart:
I'll give it another look later tonight. Until then, you two can gather up all your critiques so it can be done as a single revision rather than numerous minor edits.

gsisso:
I'll take a closer look at it tommorow,there must be someway to reduce the amount of code.
for now I can say that both getItemBall methods seem to be doing the exact same thing when strictBallType is true

LoungeKatt:
One is with probability, one isn't. The code reduction will come when the deprecated methods are removed. 
Async can call the standard, but there are much better places to reduce code than where it was just cut in half. It seems like a lot of nitpicking when half the library is still in need of optimization.

AbandonedCart:
There are two cases, one of which includes probability. There is no way to effectively accommodate that difference in the same method without somehow breaking down the login behind them.

The Async and standard options are separated to present the user with a single unified command that is only differentiated by the class of option they send to it. The entire purpose was to both cut down the redundant functions **and** to simplify the code for the end user. 

So far, many of these improvements are aimed at simply cutting lines with no real benefit to the user. I would appreciate more attention being paid to how to improve the functionality, not just imply the manner in which it was originally written was incompetent.

PR #643, opened 2016-08-19
Fill the whole signature

iGio90:
Here is my way, discussed a lot with the other guys, to fill in the whole signature. Sensor info and Location fixes has randomed (good? feel free to edit) in a good range. Sensor infos are supposed to be faked with the device hold on your hands (just like you are walking in the street and playing). Location fixes is built in a very sick way, but this is the best way i found according to the analysis of all my data. Activity status is pretty simple and for last, but not for last, the device info. I literaly change the whole approach that has been used till now. With this implementation, you will now have to feed the constructor with the unique device id (randomized) that will also be used to feed the whole device info (thanks @FabianTerhorst). This is the best way we found (optional, i leave the default constructor but it's not suggested because it will send an empty device info), since the other possible way (using the username to internally feed the device id) can't be applied due to the very first request (getPlayer) that require the signature, and that will give us the needed username (basically, the first request would have the signature without a device info).


AbandonedCart:
Why remove the option to set data and the default iPhone data to replace it with hardcoded data limited to a single device? Wouldn't this be better as an alternative or fallback?

jabbink:
:-1:

iGio90:
Edit, just found them! gonna re-add them back and rework it!

Still, DeviceInfos is the unique field it is filling right? SensorInfo still need to be filled?

@LoungeKatt agreed for fallback since with this I'm passing an unique ID instead of a random one for each session.

FabianTerhorst:
That's the wrong way. You should just find out what Sensor Infos are setted.

FabianTerhorst:
Don't destroy the current logic.

FabianTerhorst:
there are much better ways to generate the id.

iGio90:
@FabianTerhorst So, Im logging the device infos passed with the library without this commit and they are always null. I assume you are set them on your custom application elsewhere? I can see they are written in the DeviceInfo class, but I can't find any reference to it. They are just always null. In any case, I set a fallback into them if the id im generating fail. 

"There are much better ways to generate the id." in example?

FabianTerhorst:
Api.setDeviceInfo()

FabianTerhorst:
DeviceInfo default is using the uuid object.

iGio90:
That's it, so you have to call the setDeviceInfo() in your app (that implement the library)? 
Edit: DeviceInfo default is in any case using a random UUID. That point to my consideration (point 4)

FabianTerhorst:
You can save the generated device Info in the storage and Reuse it.

jabbink:
Or create it deterministically based on username + other static info

iGio90:
Allright, i tought at it. I was also think in a way to save a random one but since the library doesn't have a context itself im unable to save a random one. We could feed it with a context but im not sure it's good. Also, if you clean install back your app this uuid will change. Isn't better to build it inside the library by giving a seed?

iGio90:
@jabbink good think! i may can access with reflections to other fields like device model! we can build a fake id based on it! (edit, negative.. to much common)

FabianTerhorst:
The uuid is a seed. And you can get this back from the builded device Info.

iGio90:
@FabianTerhorst as @jabbink said, maybe we can user the username or some other static fields!

FabianTerhorst:
The random object gets the uuid.

iGio90:
@FabianTerhorst i'll try to build it with the username on the system prop

FabianTerhorst:
I will look into a way to generate a good seed for the user when I'm back in my laptop in a few hours.

iGio90:
In a few hours i hope to already have it

FabianTerhorst:
Im not sure if this fits in the library. There is nothing Bad on a different device on every Start.

mjmjelde:
Also just mentioning that SensorInfo is never passed on Android.  Or at least I have yet to see one on both the devices I am gathering data from

iGio90:
@mjmfighter 

  gps_info:
   { satellites_prn: [],
     snr: [],
     azimuth: [],
     elevation: [],
     has_almanac: [],
     has_ephemeris: [],
     used_in_fix: [] },
  sensor_info:
   { timestamp_snapshot: '2390',
     gyroscope_raw_x: 0.00830078125,
     gyroscope_raw_y: -0.0096588134765625,
     gyroscope_raw_z: -0.0039520263671875,
     accelerometer_axes: '3' },
  device_info:
   { device_id: '',
     android_board_name: 'shamu',
     android_bootloader: 'moto-apq8084-71.21',
     device_brand: 'google',
     device_model: 'shamu',
     device_model_identifier: 'MOB30O_Stock+_By_tupac4u',
     device_model_boot: 'shamu',
     hardware_manufacturer: 'motorola',
     hardware_model: 'Nexus 6',
     firmware_brand: 'shamu',
     firmware_tags: 'release-keys',
     firmware_type: 'user',
     firmware_fingerprint: 'google/shamu/shamu:6.0.1/MOB30O/2920157:user/release-keys' },

mjmjelde:
That isn't SensorInfo, thats GPSInfo.... It's different

mjmjelde:
Oh wait I see it now, I don't get that on my devices

FabianTerhorst:
Someone told me that he got some sensorinfos setted in iOS.

iGio90:
i will move to fake the sensor infos once i find a seed to generate an unique id for the device id. In my opinion, this is still wort to do

FabianTerhorst:
It's possible to integrate this into an App becuase the device Infos are settable.

mjmjelde:
@iGio90 Do you have AR enabled?  I wonder if it is set when that is enabled maybe?

iGio90:
@FabianTerhorst i dont want them settable. I want them generated :P is that so bad?

FabianTerhorst:
There are generated. When you use the defaults.

FabianTerhorst:
Some names will have the same seed. Because a string Hash isn't unique.

iGio90:
@FabianTerhorst im reinstalling my app 200 times a day :D and i want to make life easier to whose use this library because i discoverd 5 mins ago that my app was sending null device infos because of this

iGio90:
@FabianTerhorst i'll try to find a good one

FabianTerhorst:
You just have to set them.

iGio90:
anyway... since this could be discussed a lot, i think we should move the convo to slack?

FabianTerhorst:
Yes.

FabianTerhorst:
We can close this pr because these changes are not that what we realy want.

iGio90:
don't! im just reworking them and i'll also start to fill all the others! i'll let you set them from your app dont worry :D

FabianTerhorst:
Btw you know that this is using the GNU 3.0 license. And your App is violating this license.

iGio90:
Mhh not really O.o we are hosting it on bb

FabianTerhorst:
You are not providing the source.

iGio90:
But ermh.. if you need infos on this or on any other things not related to the commit, can we use slack?

FabianTerhorst:
I can use slack when I'm back on Laptop

iGio90:
Ok! i will focus now on finding a good seed, we can discuss about anything (and also on the reason for this commit of course) when you are with your baby

iGio90:
Ok awesome. I can reach any field (incl hardware ones) with reflection. It won't be hard to generate an unique id!

FabianTerhorst:
this is a java library not android.

FabianTerhorst:
so we wouldn´t add reflection to android os classes on the core module.

iGio90:
SensorInfo dumps:
https://gist.github.com/iGio90/bf978d2cfca792126d6f92e53df2d05c

FabianTerhorst:
from ios or android?

iGio90:
android

FabianTerhorst:
android filling some values different then ios.

FabianTerhorst:
we need ios network dumps.

iGio90:
I cant provide ios dumps since i doesn't own one. What I could do, instead, is using a similar default device info (with uuid that you have to save somewhere) but with android random infos?

FabianTerhorst:
This could cause a ban from niantic.

iGio90:
@FabianTerhorst i finally have tons of ios dumps ready for you :D

AbandonedCart:
You made a lot of progress. This looks pretty solid. Nice work

iGio90:
Not yet! activitystatus is fine, regarding sensor info, they are randomed good on a good range but still can happens that the values are "unnatural". Im on holydays now for 1 week but I'll keep my work on this. I want to analyze better my dumps and check if i can make them more "natural"

jabbink:
I suggest not sharing dumps with people you don't trust. Raw dumps contain your device your which niantic can ban on. And seeing as niantic has close ties to Google I wouldn't be surprised if they might even store all sensor data, so then even a specific GPS location or sensor data packet can be traced back to the account you violated the ToS with.

iGio90:
Im not playing poGo. They can ban my deviceId and whatever they want :D thanks for the interest in any case! If you need more of them just PM :) guess sensor infos wont tell nothing, i'll take it down the same to avoid problems to you all

FabianTerhorst:
I know many stuff to improve when you finished the logic for the sensorinfos.

iGio90:
Git doesnt allow you to commit on this PR?

FabianTerhorst:
when you finish this pr i will pull this pr

FabianTerhorst:
the pulled pr will be located on a branch inside this repository.

FabianTerhorst:
just focus on the logic you don´t have to resolve the conflcit.

iGio90:
Aw! Got you! I think im gonna do it this night! Sorry for the delay but my holydays started yesterday!

iGio90:
Actually i think there are no conflict O.o! I just have to make the randomed data more natural and if needed, move it to a separate class

AbandonedCart:
@FabianTerhorst https://help.github.com/articles/checking-out-pull-requests-locally/ shows how to pull the PR and as a collaborator you should also be able to push directly to it.

iGio90:
No wait, you misunderstood me... i know how to commit on my PR. I was saying that YOU can commit on this PR any time you wanna add something more or edit some of my work because you think it would be better in your way 👯

AbandonedCart:
Obviously the author of a PR can commit to it. That was for @FabianTerhorst  to append the PR.

Github for Android doesn't like tagging people when posting comments.

FabianTerhorst:
@LoungeKatt im always using this for review pull requests. This is creating a new branch locally with the commits from the author.

iGio90:
A question for you all. Do you think that raw data from sensor info are different betweens android and ios? (I.E if i put an iOS and an Android in the exact same position (assuming vertical), would the data from magne, gyro etc be the same or aprox the same?)

FabianTerhorst:
There should be the same. But android don´t fill the same values as ios.

iGio90:
I know, but since i only own an android and im trying to randomize a vertical device positioning, i want to correlate data between iOS dumps and Android ones

iGio90:
Do you think is good if i let it random on a range that suppose that your device is vertical? i dumped some data with my phone on a normal standing just like when you walk on the street and you play

FabianTerhorst:
the official app only works vertical.

iGio90:
and also, do you think there is any mod on this:

```
.addRequestEnvelopeHandler (data) ->
    buffer = pcrypt.decrypt data.unknown6?.unknown2?.encrypted_signature
    console.log POGOProtos.parseWithUnknown buffer, 'POGOProtos.Networking.Envelopes.Signature'
    false
```

so that it will only print the sensor_info message?

iGio90:
@FabianTerhorst why you always misunderstood me.... :S im not speakin about portrait or landscape. Im speaking about the device in ur hands holded in a vertical position and not dropped on a table or something else

FabianTerhorst:
I don´t know.

FabianTerhorst:
I think any possible random position is acceptable.

iGio90:
@FabianTerhorst i wanna random them in a range in which the device looks like holded on hands just like when you play on the street. The range i set in my previous commit random them on all possible position (i.e raw_x could be of the device on a desk and raw_y with the device flipped etc. that's supposed to be scriptable detected and it's not what i want..)

FabianTerhorst:
i don´t think you get banned when the screen is looking to the ground.

FabianTerhorst:
just randomice the vales in the possible ranges.

iGio90:
that's what is actually doing :D

FabianTerhorst:
But you don´t have to respect the ranges for holding the device in hand.

iGio90:
Why not?!? these range should be best!

FabianTerhorst:
you think niantic is checking if you hold the device in hands?

FabianTerhorst:
but we can use this ranges.

iGio90:
starting now the location fix. All the loves to @jabbink

iGio90:
@LoungeKatt @FabianTerhorst can you check my sick way to build the snapshot? XD

FabianTerhorst:
you don´t commited it.

iGio90:
sorry -_-

iGio90:
all other things should be easy to fake, all of this commit is just to determine a random number of providers + a random number of provider that use the negative int during the first getPlayer. From my over 2000 dumps, only the ones that start with 8 providers got 3/2 negative snapshot, all others got 5/6 providers with only 1 negative int on the first or on the second provider (almost 95% chance on the first one)

iGio90:
and everything should be natural randomed as i wrote.. maybe its a bit rough but it should reach the score :D

iGio90:
@FabianTerhorst @LoungeKatt testing now! i will report here

iGio90:
@FabianTerhorst can you confirm? dumps from my app

https://gist.github.com/iGio90/c0abf08120536a7131d6565448893a2d

iGio90:
loc fix snapshot need a little check.. something wrong there

Paul776:
What's the state of this PR? is it 'usable' ?

iGio90:
Usable

Paul776:
Sorry Haven't read all the threads. Thanks a lot for this PR! 
 people commenting here seems to be against the merge because it's not a "Android lib" ??? :/

Paul776:
Why are we using iPhone stuff?

iGio90:
Because they are more easy to spoof ^^

Paul776:
True but... isn'y that super easy to detect? Code making request is android but value are iPhones?
Is there a way from the server to tell if the request is coming from android vs iPhone device ? (any field a part from the signature?, any field in the http stuff itself?)

Do we have a dump of real request with real values ? (iPhone and android) to compare that your PR is correct?

Isn't better to set all these (at least the sensor and device) in the PokemonGO constructor? 
SO that the client can retrieve these values and store them to be able to set them back later (to avoid  having different value each time new PokemonGO is created)

iGio90:
All of the fields of my signature impl are from an iOS. Im not that sure that they can detect from which os is coming the call

iGio90:
I may have ti check the headers with charles

library/src/main/java/com/pokegoapi/util/Signature.java
```diff
@@ -8,116 +8,300 @@
...
-		StreamingXXHash64 xx64 = factory.newStreamingHash64(0x1B845238);
-		xx64.update(authTicket, 0, authTicket.length);
-		xx64 = factory.newStreamingHash64(xx64.getValue());
-		xx64.update(request, 0, request.length);
-		return xx64.getValue();
-	}
-}
+    private static Random sRandom = new Random();
+
+    private static boolean sFirstSensorInfo = true;
```
Paul776: bad

library/src/main/java/com/pokegoapi/util/Signature.java
```diff
@@ -8,116 +8,300 @@
...
+    private static boolean sFirstLocationFix = true;
```
Paul776: bad

library/src/main/java/com/pokegoapi/util/Signature.java
Paul776: Why isn't this done in PokemonGO constructor?

iGio90:
Sensor info keep changes. You are right about the device info! I'll build it only once! Why the "bad"? Can you motivate?

library/src/main/java/com/pokegoapi/util/Signature.java
Paul776: isn't there a smarter way of only setting it once without that static stuff?

Paul776:
uff didn't see the whole Signature was static :/ well leave it like this then

iGio90:
Yep, but the deviceinfos are still generated on every call and could be optimized!

Paul776:
"Sensor info keep changes"
So is it done only once for all the requests? should it be changing for each request / each accounts

iGio90:
Actually everything you see, it is generated on each request! Including deviceinfos that could be generated only once :)

iGio90:
Thanks for pointing me into it

iGio90:
Static doesnt mean that couldn't be overwritten :D

Paul776:
but because the Signature is static. it means it is shared by all the PokemonGO instances. meaning you will only get one and only one sensor and locationfix no ?

iGio90:
"static, doesn't mean it couldn't be overwritten". Do you see any if (signature != null)  :D

iGio90:
The static boolean is needed to detect if the first request has been sent

Paul776:
But because Signature is static and the boolean is static and that you have the (firstSensor first location check) they 'wont' change anymore for that PokemonGO instance no ? is that wanted?

iGio90:
Mhhhhh if you create a new istance of the api they wont change. So you are probably right :D

iGio90:
I mean, is good that they keep false status for the SAME istance. But i guess it keep it also for a new istance of pogo api

Paul776:
But ideally 
- the deviceInfo (done only once in constructor)
- sensore change at each request
- location change at each request

Maybe i don't get sometimes obvious but why is these boolean check necessary? you have to build at each request any way no?

(aka this why you should stay away from static stuff :/ not very OOP)

iGio90:
Yes, but the first time (the first request) we send sensorinfo and locfix they have stripped data! This is why i put these boolean

iGio90:
You can see it when I build the sensor infos or loc fix. The first time i send them they got stripped data

FabianTerhorst:
@Paul776 you don´t have to comment the code quality on this pr. We are using this to find the right logic. #650 is for the code.

Paul776:
@iGio90 got it

iGio90:
@Paul776 gonna mitm again now

iGio90:
@Paul776 @LoungeKatt @FabianTerhorst 
https://gist.github.com/iGio90/59d6d182167803f76afd9cdb2bd6dee5

iGio90:
Ok, explored a bit and looks really good! can you confirm?

Paul776:
{ device_id: 'H3IxFnrBavhlkqScfYUZzL5j5D3B3f0K',

"Every iPhone, iPod touch and iPad has a unique identifier number associated with it, known as a UDID (Unique Device ID). Your UDID is a 40-digit sequence of letters and numbers that looks like this: 0e83ff56a12a9cf0c7290cbb08ab6752181fb54b"

the one you set is neither 40 and has capital letters 
Can someone confirm ?

iGio90:
they are not using udid! wait im attaching you the one they are using... @LoungeKatt posted the link this morning

iGio90:
loungekatt [3:04 AM]  
it's the advertising ID. It's the new iphone udid

iGio90:
http://stackoverflow.com/questions/15802022/length-of-vendor-and-advertising-identifiers

iGio90:
mhhhh we are not sending unknown25?

Paul776:
thanks for the link. I don't have the original request (the one sent by the actual game, do you have that somewhere?)

https://possiblemobile.com/2013/04/unique-identifiers/

In the Advertiser ID section, the example ID doesn't look like the one you send (it has dashes in it)

edit: ok I'm reading that without dashes is also valid but I'm go to sleep in peace  if I see that the original request matches the one you attached :D

iGio90:
i got both ios and android dumps! but you should reach me on slack for them :D

Paul776:
aah good. if you have and compared that's cool then!

AbandonedCart:
@Paul776 The link explains it's 36 characters including 4 dashes but the iPhone dumps are returning a 32 character value, implying they parsed out the dashes. It's the only "official" udid iPhone has now that Apple restricted the use of the original udid and removed the call to it after iOS 5.

iGio90:
@LoungeKatt would you mind on telling me why we aren't sending unk25? O.O

Paul776:
@LoungeKatt got it. I just checked the dump and that's fine now

AbandonedCart:
@iGio90 I don't know any more than you about that. There are no secret meetings.

iGio90:
if (select unk25 from users == null) { fuckOff(); } xD

Paul776:
When the signature was first introduced in the API i swear i saw somewhere the setUnknown25 thing must have been removed in one of the cleanup pr? 

Edit: naaa sorry it was unk22

iGio90:
nope... it has never been built... and it's static.................

Paul776:
Do we know what it is?

PR #645, opened 2016-08-20
Improving Login Stability

yaohongkok:
**Fixed issue:** Login is not stable (Sometimes, my app can login into the server)

**Changes made:**
- Added sleep(800) to the constructor
- Added throw InterruptException

My friends and I use this API to do some work of our own. We noticed
that login is not stable. I suggest to add some Sleep in this code. It
did made my login more stable.


AbandonedCart:
This should probably be a parameter to allow configuring your own sleep value. It's too unpredictable to set a predefined, since it depends on network speed, server load, etc.

You also need to include the imports and javadocs

yaohongkok:
Sure. I will add the parameter into it. You have a good point. I am on a
fast network...

Thanks!
ᐧ

yaohongkok:
Due to some feedback, I am closing this pull request so that I can fix the issues.

PR #647, opened 2016-08-21
Improve login stability

yaohongkok:
**Fixed issue:** Improved PokemonGo.java constructor (and hence login) stability

**Changes made:**
- Added sleep after PlayerProfile() and Settings() because they request info from the server. From some previous issues, we know that we need some time delay between request.
- The sleep will only be turned on after adding a "true" argument to the constructor so that it doesn't break existing code
- Parametrized the sleep time

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/PokemonGo.java
+++ b/library/src/main/java/com/pokegoapi/api/PokemonGo.java
@@ -41,6 +41,7 @@ import java.util.Random;
 public class PokemonGo {
 
 	private static final java.lang.String TAG = PokemonGo.class.getSimpleName();
+	private static final int sleepTime = 800;
 	private final Time time;
 	public final long startTime;
 	@Getter
@@ -72,11 +73,13 @@ public class PokemonGo {
 	 * @param credentialProvider the credential provider
 	 * @param client             the http client
 	 * @param time               a time implementation
+	 * @param needSleep			 need sleep after creating settings
 	 * @throws LoginFailedException  When login fails
 	 * @throws RemoteServerException When server fails
+	 * @throws InterruptedException When Thread.sleep fails
 	 */
 
-	public PokemonGo(CredentialProvider credentialProvider, OkHttpClient client, Time time)
+	public PokemonGo(CredentialProvider credentialProvider, OkHttpClient client, Time time, boolean needSleep)
 			throws LoginFailedException, RemoteServerException, InterruptedException {
 
 		if (credentialProvider == null) {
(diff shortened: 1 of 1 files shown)


yaohongkok:
Please refer to my updated code as to why there are 2 sleeps. :)

gionata-bisciari:
I had already read the commit info

gionata-bisciari:
```
/home/travis/build/Grover-c13/PokeGOAPI-Java/sample/src/main/java/com/pokegoapi/examples/CatchPokemonAtAreaExample.java:63: error: unreported exception InterruptedException; must be caught or declared to be thrown
            PokemonGo go = new PokemonGo(new PtcCredentialProvider(http, ExampleLoginDetails.LOGIN,
                           ^
/home/travis/build/Grover-c13/PokeGOAPI-Java/sample/src/main/java/com/pokegoapi/examples/UseIncenseExample.java:53: error: incompatible types: SystemTimeImpl cannot be converted to boolean
            PokemonGo go = new PokemonGo(authProvider, http, new SystemTimeImpl());
```

AbandonedCart:
This is how I am using it: https://github.com/AbandonedCart/PokeGOAPI-Java/commit/1c054d1a89a56560f69add4e0a6ccf97a1314afe if that helps.

yaohongkok:
Hi everyone,

Instead of throwing the exception, I will surround it with try and catch
block. Sorry about the issue.

Will fix it soon.

ᐧ

yaohongkok:
Actually, TwistedUmbrella's method is actually pretty good. I might change
it to a delay instead. Thanks for pointing it out.
ᐧ

library/src/main/java/com/pokegoapi/api/PokemonGo.java
```diff
@@ -72,12 +73,14 @@
 	 * @param credentialProvider the credential provider
 	 * @param client             the http client
 	 * @param time               a time implementation
+	 * @param needSleep			 need sleep after creating settings
 	 * @throws LoginFailedException  When login fails
 	 * @throws RemoteServerException When server fails
+	 * @throws InterruptedException When Thread.sleep fails
```
iphydf, line 79: `Thread.sleep` doesn't fail. `InterruptedException` happens when another thread interrupts the current one. Although Java likes to encode unexceptional situations in exceptions, this is not a failure mode.

FabianTerhorst, line 79: take a look at #654 please

AbandonedCart:
The default should probably be to use it and the purpose of the one I was demonstrating was to configure the actual time, not just make it an optional parameter altogether. It should only be an arbitrary 800 if you have done extensive tests to verify 800 is the exact required value.

@FabianTerhorst PRs should never account for other PRs until they are merged.

yaohongkok:
Hi everyone,

Your feedback are greatly appreciated. I will incorporate them today or
tomorrow. Then, I will ask you guys to give it another check.
ᐧ

FabianTerhorst:
We will combine the PlayerProfile, Inventories... request into one request array in part 5 of the login rework.

FabianTerhorst:
I don´t think we will need this after that anymore

AbandonedCart:
This has been overridden by the current changes to the login.

PR #650, opened 2016-08-22
Improve signature with activity status, improved device- and sensor info

FabianTerhorst:
This pr contains the logical changes from #643.

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/PokemonGo.java
+++ b/library/src/main/java/com/pokegoapi/api/PokemonGo.java
@@ -69,12 +69,12 @@ public class PokemonGo {
 	private Map map;
 	@Setter
 	private DeviceInfo deviceInfo;
-	@Setter
-	@Getter
-	private long seed;
 	@Getter
 	@Setter
 	public SensorInfo sensorInfo;
+	@Setter
+	@Getter
+	private long seed;
 	@Getter
 	@Setter
 	public List<SignatureOuterClass.Signature.LocationFix> locationFixes;
--- a/library/src/main/java/com/pokegoapi/api/PokemonGo.java
+++ b/library/src/main/java/com/pokegoapi/api/PokemonGo.java
@@ -19,6 +19,7 @@ import POGOProtos.Networking.Envelopes.RequestEnvelopeOuterClass.RequestEnvelope
 import POGOProtos.Networking.Envelopes.SignatureOuterClass;
 
 import com.pokegoapi.api.device.DeviceInfo;
+import com.pokegoapi.api.device.LocationFixes;
 import com.pokegoapi.api.device.SensorInfo;
 import com.pokegoapi.api.inventory.Inventories;
 import com.pokegoapi.api.map.Map;
@@ -74,7 +75,7 @@ public class PokemonGo {
 	private long seed;
 	@Getter
 	@Setter
-	public List<SignatureOuterClass.Signature.LocationFix> locationFixes;
+	public LocationFixes locationFixes;
 
 	/**
 	 * Instantiates a new Pokemon go.
--- a/library/src/main/java/com/pokegoapi/api/PokemonGo.java
+++ b/library/src/main/java/com/pokegoapi/api/PokemonGo.java
@@ -18,9 +18,7 @@ package com.pokegoapi.api;
 import POGOProtos.Networking.Envelopes.RequestEnvelopeOuterClass.RequestEnvelope.AuthInfo;
 import POGOProtos.Networking.Envelopes.SignatureOuterClass;
 
-import com.pokegoapi.api.device.ActivityStatus;
 import com.pokegoapi.api.device.DeviceInfo;
-import com.pokegoapi.api.device.LocationFix;
 import com.pokegoapi.api.device.SensorInfo;
 import com.pokegoapi.api.inventory.Inventories;
 import com.pokegoapi.api.map.Map;
@@ -37,7 +35,6 @@ import lombok.Getter;
 import lombok.Setter;
 import okhttp3.OkHttpClient;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Random;
 import java.util.UUID;
--- a/library/src/main/java/com/pokegoapi/api/PokemonGo.java
+++ b/library/src/main/java/com/pokegoapi/api/PokemonGo.java
@@ -250,6 +250,6 @@ public class PokemonGo {
 	 * @return the activity status
 	 */
 	public SignatureOuterClass.Signature.ActivityStatus getActivityStatus() {
-		return ActivityStatus.getDefault(this);
+		return ActivityStatus.getDefault();
 	}
 }
--- a/library/src/main/java/com/pokegoapi/api/PokemonGo.java
+++ b/library/src/main/java/com/pokegoapi/api/PokemonGo.java
@@ -89,7 +89,7 @@ public class PokemonGo {
 
 	public PokemonGo(CredentialProvider credentialProvider, OkHttpClient client, Time time, long seed)
 			throws LoginFailedException, RemoteServerException {
-
+		startTime = currentTimeMillis();
 		if (credentialProvider == null) {
 			throw new LoginFailedException("Credential Provider is null");
 		} else {
@@ -107,7 +107,6 @@ public class PokemonGo {
 		map = new Map(this);
 		longitude = Double.NaN;
 		latitude = Double.NaN;
-		startTime = currentTimeMillis();
 	}
 
 	/**
@@ -137,7 +136,7 @@ public class PokemonGo {
 	 */
 	public PokemonGo(CredentialProvider credentialProvider, OkHttpClient client, Time time)
 			throws LoginFailedException, RemoteServerException {
-		this(credentialProvider, client, time, UUID.randomUUID().hashCode());
+		this(credentialProvider, client, time, hash(UUID.randomUUID().toString()));
 	}
 
 	/**
@@ -151,7 +150,29 @@ public class PokemonGo {
 	 */
 	public PokemonGo(CredentialProvider credentialProvider, OkHttpClient client)
 			throws LoginFailedException, RemoteServerException {
-		this(credentialProvider, client, new SystemTimeImpl(), UUID.randomUUID().hashCode());
+		this(credentialProvider, client, new SystemTimeImpl(), hash(UUID.randomUUID().toString()));
+	}
+
+	/**
+	 * @param string string to hash
+	 * @return the hashed long
+	 */
+	private static long hash(String string) {
+		long upper = ((long) string.hashCode()) << 32;
+		long lower = ((long) reverse(string).hashCode()) - ((long) Integer.MIN_VALUE);
+		long hash64 = upper + lower;
+		return hash64;
+	}
+
+	private static String reverse(String string) {
+		int i, len = string.length();
+		StringBuilder dest = new StringBuilder(len);
+
+		for (i = (len - 1); i >= 0; i--) {
+			dest.append(string.charAt(i));
+		}
+
+		return dest.toString();
 	}
 
 	/**
--- a/library/src/main/java/com/pokegoapi/api/PokemonGo.java
+++ b/library/src/main/java/com/pokegoapi/api/PokemonGo.java
@@ -160,16 +160,15 @@ public class PokemonGo {
 	private static long hash(String string) {
 		long upper = ((long) string.hashCode()) << 32;
 		long lower = ((long) reverse(string).hashCode()) - ((long) Integer.MIN_VALUE);
-		long hash64 = upper + lower;
-		return hash64;
+		return upper + lower;
 	}
 
 	private static String reverse(String string) {
-		int i, len = string.length();
+		int index, len = string.length();
 		StringBuilder dest = new StringBuilder(len);
 
-		for (i = (len - 1); i >= 0; i--) {
-			dest.append(string.charAt(i));
+		for (index = (len - 1); index >= 0; index--) {
+			dest.append(string.charAt(index));
 		}
 
 		return dest.toString();
--- a/library/src/main/java/com/pokegoapi/api/PokemonGo.java
+++ b/library/src/main/java/com/pokegoapi/api/PokemonGo.java
@@ -20,6 +20,7 @@ import POGOProtos.Networking.Envelopes.SignatureOuterClass;
 
 import com.pokegoapi.api.device.ActivityStatus;
 import com.pokegoapi.api.device.DeviceInfo;
+import com.pokegoapi.api.device.LocationFix;
 import com.pokegoapi.api.device.SensorInfo;
 import com.pokegoapi.api.inventory.Inventories;
 import com.pokegoapi.api.map.Map;
@@ -36,6 +37,7 @@ import lombok.Getter;
 import lombok.Setter;
 import okhttp3.OkHttpClient;
 
+import java.util.List;
 import java.util.Random;
 import java.util.UUID;
 
@@ -68,6 +70,12 @@ public class PokemonGo {
 	private DeviceInfo deviceInfo;
 	@Setter
 	private String uuid;
+	@Getter
+	@Setter
+	private boolean firstSensorInfo;
+	@Getter
+	@Setter
+	private boolean firstLocationFix;
 
 	/**
 	 * Instantiates a new Pokemon go.
@@ -237,5 +245,14 @@ public class PokemonGo {
 	 */
 	public SignatureOuterClass.Signature.ActivityStatus getActivityStatus() {
 		return ActivityStatus.getDefault();
+	}
+
+	/**
+	 * Gets the location fix
+	 *
+	 * @return the sensor info
+	 */
+	public List<SignatureOuterClass.Signature.LocationFix> getLocationFix() {
+		return LocationFix.getDefault(this);
 	}
 }
--- a/library/src/main/java/com/pokegoapi/api/PokemonGo.java
+++ b/library/src/main/java/com/pokegoapi/api/PokemonGo.java
@@ -67,16 +67,9 @@ public class PokemonGo {
 	@Setter
 	private DeviceInfo deviceInfo;
 	@Setter
-	private SensorInfo sensorInfo;
-	@Setter
-	private ActivityStatus activityStatus;
-	@Setter
 	private String uuid;
 	@Setter
 	private Random random;
-	@Getter
-	@Setter
-	private boolean firstSensorInfo;
 
 	/**
 	 * Instantiates a new Pokemon go.
@@ -248,10 +241,7 @@ public class PokemonGo {
 	 * @return the sensor info
 	 */
 	public SignatureOuterClass.Signature.SensorInfo getSensorInfo() {
-		if (sensorInfo == null) {
-			sensorInfo = SensorInfo.getDefault(this);
-		}
-		return sensorInfo.getSensorInfo();
+		return SensorInfo.getDefault(this);
 	}
 
 	/**
@@ -260,9 +250,6 @@ public class PokemonGo {
 	 * @return the activity status
 	 */
 	public SignatureOuterClass.Signature.ActivityStatus getActivityStatus() {
-		if (activityStatus == null) {
-			activityStatus = ActivityStatus.getDefault(this);
-		}
-		return activityStatus.getActivityStatus();
+		return ActivityStatus.getDefault(this);
 	}
 }
--- a/library/src/main/java/com/pokegoapi/api/PokemonGo.java
+++ b/library/src/main/java/com/pokegoapi/api/PokemonGo.java
@@ -78,6 +78,9 @@ public class PokemonGo {
 	@Getter
 	@Setter
 	public List<SignatureOuterClass.Signature.LocationFix> locationFixes;
+	@Getter
+	@Setter
+	private String versionCode;
 
 	/**
 	 * Instantiates a new Pokemon go.
@@ -111,6 +114,8 @@ public class PokemonGo {
 		longitude = Double.NaN;
 		latitude = Double.NaN;
 		startTime = currentTimeMillis();
+
+		versionCode = "2016080700";
 	}
 
 	/**
--- a/library/src/main/java/com/pokegoapi/api/PokemonGo.java
+++ b/library/src/main/java/com/pokegoapi/api/PokemonGo.java
@@ -18,6 +18,7 @@ package com.pokegoapi.api;
 import POGOProtos.Networking.Envelopes.RequestEnvelopeOuterClass.RequestEnvelope.AuthInfo;
 import POGOProtos.Networking.Envelopes.SignatureOuterClass;
 
+import com.pokegoapi.api.device.ActivityStatus;
 import com.pokegoapi.api.device.DeviceInfo;
 import com.pokegoapi.api.device.LocationFixes;
 import com.pokegoapi.api.device.SensorInfo;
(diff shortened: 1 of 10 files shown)


iGio90:
🎉

FabianTerhorst:
you can commit more to #643 will add this here.

iGio90:
@FabianTerhorst im doing location fix :) it will be out today! gf birthday today.. she is about to kill me

AbandonedCart:
Does this invalidate the need to even call these items when using the default?

FabianTerhorst:
You only have to set them when you Relay want your own.

AbandonedCart:
So then yes. That's much better than the current design of having to arbitrarily set "DeviceInfo.DEFAULT"

iGio90:
Do this generate by default a random device id every time it's initialized?? I mean, it will change every time right?

FabianTerhorst:
When you look into the code you can See that you can Set the uuid and or the random object.

iGio90:
Yep I saw it, but i will make a new PR later when this is merged to auto generate one using the username as seed if the app whose implement doesn't provide anything on it

AbandonedCart:
It should NOT set any random as part of the constructors. It should be set outside and cached. You shouldn't be changing devices and IDs every time you move. With that, it only sets uuid as a new random if the items is null, but the device is set as a new one each time you call a default device.

https://github.com/Grover-c13/PokeGOAPI-Java/pull/650/commits/275c923811c7a724d393d95b1644bd32895e2e01#diff-b640d379a8fad8de11ef3c425c809a56R175

FabianTerhorst:
The Default device will be only created once. Because it's get Stored inside the api instance.

AbandonedCart:
Then is there a real need to expose the UUID generation? Or more appropriately, the random generated from it, since that is only being used to fill in the device info?

library/src/main/java/com/pokegoapi/api/device/DeviceInfo.java
```diff
@@ -115,6 +119,70 @@ public DeviceInfo(DeviceInfos deviceInfos) {
...
 	/**
+	 * Gets the default device status for the given api
+	 *
+	 * @param api the api
+	 * @return the default device info for the given api
+	 */
+	public static DeviceInfo getDefault(PokemonGo api) {
+		DeviceInfo deviceInfo = new DeviceInfo();
+		deviceInfo.setDeviceId(api.getUuid());
+		Random random = api.getRandom();
```
AbandonedCart: could be new `Random random = new Random(api.getUuid().hashCode());` to eliminate the entire Random object in another class, since knowing the UUID may be beneficial but setting your own UUID implies you probably won't want an iPhone based on the UUID.

iGio90:
We should generate an iOS uid not an android one. It's way different. Leave this as this, i'll work on it once I finish the location fix!

AbandonedCart:
Either way, it seems proprietary so the point stands that exposing the random object generated from it will have no use outside generating the actual profile.

You could also support a parameter in the UUID generation to specify the type of device targeted (ie Android or iPhone).

FabianTerhorst:
Im setting the random object, because the seed isn´t 100% unique. You can´t convert an String to an long without loosing bytes.

FabianTerhorst:
Thats the reason why i don´t use the username as a seed.

FabianTerhorst:
I will also rename uuid to seed.

AbandonedCart:
In that case, none of it needs to be exposed then. But then it doesn't make sense to set the device ID as the seed, so it is still being used as a UUID in some places.

FabianTerhorst:
we just have to generate a random long so the user can save them and load them on next start

AbandonedCart:
I was asking what purpose there is in setting the random object externally when you have stated the entire default device is only generated once. If the device is generated, then the Random instance used to generate it is no longer relevant. You also set the value you currently call UUID as the device ID, but you are saying it is not a UUID but a seed. If it is not a UUID, there is no need to set it externally either, since it is only an instance to create the device info. If it is a UUID, then it may be useful to retrieve it, but more appropriately placed in the device info than the entire api object.

iGio90:
@FabianTerhorst the seeds is unique if you use the login username either ptc or google :)

FabianTerhorst:
How do you want to get an String into an long without loosing bytes?

iGio90:
im not understanding what you mean. The uuid you are generating is actually a string. In the same way we can build a very simple string generator that will use the username as seed

AbandonedCart:
The device UUID is the DEVICE ID. A seed for generating device info is a seed for generating device info. The two are not the same thing. Seems there is some confusion by labeling them both a UUID. 
http://stackoverflow.com/questions/5088474/how-can-i-get-the-uuid-of-my-android-phone-in-an-application <- A device ID
http://stackoverflow.com/questions/21872233/differences-between-udid-and-uuid <- A device ID
https://developer.android.com/reference/java/util/UUID.html <- A generated UUID

iGio90:
Im labeling uuid the device id not the device info. After I build the device id with the username as seed, i'll use the device id to obtain an unique device info (so it will always be iOS 6 or 7 or 8 etc etc) hope it is more understable. Can you leave this device info now as i will take it this night! checkout the sensor info if it's ok for you by reading the code, im testing now the values returned if they are ok after a quick coffe + cig 💃

AbandonedCart:
What you may be doing is not what @FabianTerhorst is describing or this code is doing. This code it generating a UUID, setting it to the device ID, then using that as a seed to generate the info.
Theoretically, that is the most efficient way to do it, but it comes back to the random object not being necessary outside of generating the device info.

FabianTerhorst:
the seed for a random is a long

FabianTerhorst:
I will put the isFirst... SensorInof into the sensorinfo object and the uuid to the device info object.

AbandonedCart:
The type is insignificant. You are still basing it on the value used for UUID as per the code you submitted at https://github.com/Grover-c13/PokeGOAPI-Java/pull/650/commits/275c923811c7a724d393d95b1644bd32895e2e01#diff-d789d3aec5f083a79f177a3df5e4a914R230

iGio90:
But the uuid WILL NOT be random and mustn't be random! I'm meaning, i'll use the username to build a string (alphanumeric) based on the username... guys what's the hard part of what im writing? maybe my english is bad? 👯 username is unique since it's the account which you use to login and it's fine that it change if you change the login!!!!!!!

FabianTerhorst:
the random object constructor needs a long as an seed. You can convert an string to an long but you will loose data. That means some usernames have the same seed long at the end.

AbandonedCart:
@iGio90 You aren't understanding the conversation. The UUID is generated once, and then it is no longer random. The random is generated based on that single unique value that is then stored.

It is derived from it. It is not directly it. It is the seed in a manner of speaking, but not the final seed. The final seed is the appropriate type value derived from the cached value of the UUID. <-technical

iGio90:
But why you need to feed the random with a seed? O.O this is the thing i'm missing, what is the purpose?

FabianTerhorst:
same seed = same output.

AbandonedCart:
Technically, you don't. It is just a good basis if you want to extract the UUID for future use.

iGio90:
We don't need same output except for the uuid

iGio90:
All my commits are using just random things and that's the real purpose of a "random". You just dont need the same output on further sessions

AbandonedCart:
But if you don't feed that same UUID in every time as the basis for the random, you will also get a different device every time, not the random device for that UUID. It is a red flag that you have 345 iPhones for 345 sessions, wouldn't you think?

iGio90:
why ahhaha you can just tell it, if uuid.startswith a value from a to f use device a otherwise use device b. we got tons of way to accomplish it...

AbandonedCart:
Then you cache a ton of devices when all you need is a single value.

iGio90:
"cache a tons of devices" you mean keep them in an arraylist as @FabianTerhorst is doing?

mjmjelde:
I feel like you guys are way overthinking a lot of this.  Device ID = 16 char hash of username.  If Random is needed for the rest of the device info, just use username.length casted to a long for the seed.  Use you might get a few repeated device if the usernames you are using are the same length, but what of it.  There are thousands of the same devices out there that are the same with just different device ids.  You don't need any special way of generating the UUID for a new user and storing it to some place for later use... username is always going to be specific to that account... why not use it.

FabianTerhorst:
Is there no userid returned by the api?

AbandonedCart:
Which goes back to my original comment of "why is UUID and random outside of generating the random device"

@FabianTerhorst if you remember, people reported the one thing they could retrieve without the new signature was the player profile and if you remember signing up, you couldn't use a username already in use. Why do you think both would be true?

They practically already admitted anything that is generated is based on those items. Make sense we would just use their own methods.

AbandonedCart:
I'm not really concerned with what seed is used in the final revision, all that really matters is not making it so random that device info is never the same for two separate references to the same instance and if there is a "seed" that isn't something I can access in the API, I can cache it to use it again in another instance and get back the same (or similar) device.

FabianTerhorst:
When we would use the username length like @mjmfighter suggested we don´t have to give him back so someone can store him

AbandonedCart:
`if there is a "seed" that isn't something I can access in the API` <- that would be covered then.

(Edit: No real reason for asking a rhetorical question)

FabianTerhorst:
sensor info and activity status will be generated every time now.

iGio90:
Check my PR. Is the best way you can do it. The first getPlayer, that will gives you the username also need the signature. You will send the first getPlayer without the deviceinfos because you still dont have the username. Your code will return an npe!

iGio90:
"is the best way you can do it" -> imo

FabianTerhorst:
The seed is a long so it cant be null. The Problem is that we do request in the constructor. That's not realy a good Pattern. So i will make it possilbe to set the seed with a constructor Argument.

iGio90:
Fabian.. i really hate that when I speak with you i have to repeat the things 20000 times. I did work on it all the night. The first signature you are going to generate. THE FIRST. You still dont have any PlayerProfile because you build it AFTER the getPlayer. So it will gives you an NPE at first and so you have to use an empty fallback string that will change ur deviceId. You cant use username, what will you use to generate the deviceInfo of the first signature?

library/src/main/java/com/pokegoapi/api/PokemonGo.java
```diff
@@ -108,6 +109,7 @@ public PokemonGo(CredentialProvider credentialProvider, OkHttpClient client, Tim
 		longitude = Double.NaN;
 		latitude = Double.NaN;
 		startTime = currentTimeMillis();
+		seed = playerProfile.getPlayerData().getUsername().length();
```
iGio90: ok, I'm at PC now. This will throw an npe at the first signature generation because getPlayerData is null

library/src/main/java/com/pokegoapi/api/PokemonGo.java
```diff
@@ -108,6 +109,7 @@ public PokemonGo(CredentialProvider credentialProvider, OkHttpClient client, Tim
 		longitude = Double.NaN;
 		latitude = Double.NaN;
 		startTime = currentTimeMillis();
```
iGio90: This need to be moved on top

library/src/main/java/com/pokegoapi/api/PokemonGo.java
Paul776: Please stop putting http calling the damn constructor

Paul776: Can we stop using http calls in the constructor ? (ignore it this is not a call)

AbandonedCart:
As it stands, this is kind of a mess and merges three different PRs into one. 
The sample is not broken, nor is it wrong for providing a demonstration that time is an accepted parameter and I have fixed the checkstyle in a separate commit so this is no longer used as a prerequisite for unrelated PRs.

Paul776:
Has someone taken this PR, MiTM'led it and did the same request with a real phone?

Are the requests 100% identical a part from the device model etc.. ? Is there any other field in the proto that are unknown and that can contain info that turn this PR useless ?

iGio90:
I did the mitm with my PR. I also got the gist if needed. I dont know if he is using my same logic, but if so, its just fine :)

Paul776:
Had a look through this PR, it looks ok to me (not 100% sure about al these values from the sensors ). I'm gonna build the the API with this PR and canary deploy it to my users. will see in couple days

FabianTerhorst:
@LoungeKatt i added the missing constructor and revert the change on the sample.

FabianTerhorst:
im using now the right device id like @jabbink is using.

iGio90:
how can it be unique? (from device to device)

FabianTerhorst:
It's using the provided seed.

FabianTerhorst:
When you provide the real deviceid from an Android device as a seed you have a unique deviceid.

iGio90:
oh yeah... i saw it now... but still, they can reverse this and get to the real device id. Better to provide a random uuid that you would save within your app instead of the real device id

FabianTerhorst:
then you shouldn't provide your real deviceid seed and get one generated. It's a very low chance to get the same deviceid twice. Even when it's twice no one cares if you use one account on two device.

iGio90:
agree with you.. this time :D

FabianTerhorst:
You can of corse save the generated ids on a dabase and give them back to the right account and generate only unique ones.

iGio90:
That's what im doing with my PR :D

iGio90:
did you check the dump btw?

FabianTerhorst:
you are not checking for unique with a external database.

iGio90:
check my last commit and please, add unknown25! why we arent sending it at all???? O.O

FabianTerhorst:
I added unknown 25. I Don´t have to replace "-" with "", because im generating the device id without the uuid object. Im only generating the seed with the uuid object.

AbandonedCart:
`int64 unknown25 = 25; // for 0.33 its static -8537042734809897855 or 0x898654dd2753a481, generated via xxHash64("\"b8fa9757195897aae92c53dbcf8a60fb3d86d745\"".ToByteArray(), 0x88533787)`

That looks like it should be xxHash64(UDID.toByteArray(), 0x88533787)

iGio90:
@LoungeKatt cant be udid.tobytearray... it's static for everyone. I think it's a string pulled from the server. I can't find it hardcoded anywhere

AbandonedCart:
How would it be a string pulled from the server to make the initial request to the server? 
If so, there needs to be a way to get the actual string or that is an easy flag.

FabianTerhorst:
It should be generated locally. Is the Value always the same on every day?

iGio90:
yes! it's static every day for every device. Just changed with 0.35

iGio90:
Can you try generate it with the version of the app? either 0.35 or 035

iGio90:
Anyway... if we really want to make everything safe and ban proff, we need also to replicate the requests that the official clean is doing. In example, we are sending the get player twice at login just like the official client is doing. But what about all the others? I.E this is fired right after login (every time)

[<-] Request for DownloadRemoteConfigVersion  { platform: 'IOS', app_version: 3500, '$unknownFields': [] } 

[<-] Request for GetHatchedEggs  {} 

[<-] Request for GetInventory  {} 

[<-] Request for CheckAwardedBadges  {} 

[<-] Request for DownloadSettings  {}

iGio90:
could it be this one? from the first getPlayer:

[<-] Request for GetPlayer  { app_version: '\n\u0002CO\u0012\u0002es',

FabianTerhorst:
Unknown 25 = app_version ?

iGio90:
Nope.
int64 unknown25 = 25; // for 0.33 its static -8537042734809897855 or 0x898654dd2753a481, generated via xxHash64("\"b8fa9757195897aae92c53dbcf8a60fb3d86d745\"".ToByteArray(), 0x88533787)

b8fa9757195897aae92c53dbcf8a60fb3d86d745 <- sha1 of something

that something could be the app version (just a guess)

AbandonedCart:
It has to be something they could calculate when building the app, otherwise it would be sent from the server to be used in future requests (which would also be a little silly, honestly).

FabianTerhorst:
Maybe the Version code and not the Version Name.

AbandonedCart:
sha1 of 0.35.0 is 642426b9fbbf0936f8799893e073fbf7688e8c82
sha1 of 0.33.0 is e66b972c1fd9132a1fff6ff6f32442ea085ca259

Anyone have the manifest for 0.33.0?

FabianTerhorst:
thats not the version code

FabianTerhorst:
thats the version code: 2016082200

AbandonedCart:
How do you prove it is not the version name? You test it. 
You also need the version code for 0.33.0 since that is the one we have a logged value for, but the one you posted is unlabeled.

AbandonedCart:
sha1 of 2016082200 is18e53dfcd47b052f28ac7354e2521b8d718b40b5

FabianTerhorst:
0.33.0 is 2016080700

AbandonedCart:
bda3f7c03cf14d9adb38aa75ad81f1aafccd25af (so close)

FabianTerhorst:
Now we know its not the version code and not the version name

AbandonedCart:
sha1 hash b8fa9757195897aae92c53dbcf8a60fb3d86d745 = Sorry, no solution found.

Paul776:
No one on the discord chans have figured out what the uk25 is ?

FabianTerhorst:
2016082200 sha1 hash = b8fa9757195897aae92c53dbcf8a60fb3d86d745 realy?

Paul776:
18E53DFCD47B052F28AC7354E2521B8D718B40B5

iGio90:
yep... told you! version code GG

Paul776:
gg gio

iGio90:
no, gg @FabianTerhorst that get it :D i just gave the idea

FabianTerhorst:
But i had never looked into the version code without you.

iGio90:
well.. gg everybody then ^^

iGio90:
Im making a PR to the proto then. How should we name it?

iGio90:
version_code?

AbandonedCart:
Mac md5 is wrong, so it's no surprise sha1 would be but that was using the sha1sum from macports. Apparently I need to file a bug report. Oh well, some good came from it at least.

AbandonedCart:
Should probably name it versionCode to match the actual variable used by Unity and friends.

iGio90:
i set it to version_code_hash. Should fit it! well... gonna get back on the signature now. After we finish, i think we have to seriously consider a rework of the login flow to pair the official client. We have a totally different one

iGio90:
in relation to unk25, we should send it as we are doing now since we have to manual change it in the same way every time an update come

LoungeKatt:
I'm submitting some documentation to the protos. Nice work guys.

AbandonedCart:
So we would still generate the final value as xxHash64(SHA1_HASH.ToByteArray(), 0x88533787), since what we discovered was the origin of the internal value in the method. That may mean unk25 isn't actually the version code hash so much as the seed.

iGio90:
@LoungeKatt yep, as I said, we should keep sending it as the final value since we have to manually insert the seed on each update

AbandonedCart:
I mean in terms of mislabeling it. It should probably be version_identifier or something less specific.

iGio90:
uint64 location_hash2 = 20; // Location2 hashed (unsigned) - xxHash32
int64 version_code_hash = 25; // version code sha1 hash - xxHash64

mhhh you are probably right but then all other fields should be renamed as well, no?

AbandonedCart:
But location2 is an object and that is the xxHash32 of it. This is a hash of the hash, which is a terrible variable name lol

iGio90:
yeah hahahah it's hard to name it! // comment will be our friend

AbandonedCart:
~~int64 hash_of_hash_of_version_code_not_unknown25~~
int64 hash_of_hash_of_version_code_not_unk25

iGio90:
_not_unknown25 is necessary? :D

AbandonedCart:
That wasn't serious. We can probably move this to slack now that it is just semantics.

iGio90:
Mhhh is it good to in your opinion to replicate the whole encryption process when we can just send the crypted value? We still have to edit the seed in any case

Paul776:
Well it's not the version code then

FabianTerhorst:
its not the version code. I thought @iGio90 and @LoungeKatt could reproduce the output, but that doesn´t looks like this right now.

AbandonedCart:
The values I posted were what I got, but I figured it was just the Mac <-> Linux issue that prevented me from getting the "correct" one.

library/src/main/java/com/pokegoapi/util/Signature.java
```diff
@@ -64,7 +49,7 @@ public static void setSignature(PokemonGo api, RequestEnvelopeOuterClass.Request
 				.setSensorInfo(SensorInfo.getDefault(api))
 				.setActivityStatus(ActivityStatus.getDefault())
 				.addAllLocationFix(LocationFix.getDefault(api))
-				.setVersionCodeHash(versionCodeHash);
+				.setVersionCodeHash(-8537042734809897855L);
```
FabianTerhorst: is this the correct hard coded value?

AbandonedCart: Why fill it in? Why not leave the functionality there to xxHash64 the "whatever it actually is" and fill that part in instead so in the future it can easily be replaced by whatever it actually is.

FabianTerhorst: because it will fill in a invalid value.

AbandonedCart: xxHash64("\"b8fa9757195897aae92c53dbcf8a60fb3d86d745\"".ToByteArray(), 0x88533787)?

FabianTerhorst: ok thats also possibe

FabianTerhorst:
ok, whats next?

AbandonedCart:
Clean it up and submit it for now, I believe

FabianTerhorst:
I think we can get rid of the interfaces for the device and for the sensor infos.

FabianTerhorst:
I have to create some new accounts for testing.

AbandonedCart:
Personally, I prefer having the interfaces for setting things to my actual device, albeit not the best idea currently. I don't know that they really need to be deprecated.

Also, when squashing you can re-order commits by cutting the entire line and pasting it elsewhere. That may help if you squash and want to move some things around the proto updates

FabianTerhorst:
I will squash when merge to squash the "merged by" commit as well.

Paul776:
@FabianTerhorst @LoungeKatt @iGio90 Have you guys checked gio's iOS dumped? sensor_info and location_fix are not send every time. Moreover sensor_info doesn't send the FULL info each time. 

I haven't figured out the pattern yet (when and when not send location_fix and when and when not send sensor_info and when and when not send the FULL sensor_info) by just reading the dump

iGio90:
@Paul776 im nearly done :) i got new dumps from my app that are same as official client. However, there is more work needed by us to make a banproff client. A reworked login is necessary + we need to add the locale to the getplayer request

AbandonedCart:
Is this part complete, at least? It seems like we have a solid base to begin working from.

iGio90:
I dont know about Fabian code. I will attach a gist on the chat after this Quick cigarette from my PR that is completed. I just need to commit the last changes

AbandonedCart:
From what I've seen, it does the same basic process while leaving open endpoints to edit the sensor, activity, and device info from the user side.

iGio90:
Nope... i changed a lot

FabianTerhorst:
I will take a look into this tomorrow(25.)

PR #652, opened 2016-08-22
clear cache when location is changed

### LINKED ISSUE #646: getCatchablePokemon() and getWildPokemon() unexpected behavior
**Description:**
When I call getCatchablePokemon() and getWildPokemon() at a certain coordinate, say (40.781549,-73.9650843646)
I get a set of pokemon, which is expected. However, when I make another right after on a point a couple of hundred meters away (40.781549,-73.9638913646) I get the exact same set of pokemon for the second point as well. 
When I only call the second point by itself I get an empty set of pokemon or a different set.
The problem occurs when I call two or more points in succession. 

**Steps to reproduce:**
1. [Step 1] Call getWildPokemon() at point A 
2. [Step 2] Call getWildPokemon() at point B (Note that i get the same set as in point A)
3. [Step 3] Call getWildPokemon() at point B (Get empty set or different pokemon)

**Expected behavior:**
I should be getting a different set of pokemon per point, provided that the points don't over lap.

**Actual behavior:**
I get the same set of pokemon for both points, but different sets if called separetly. 

**Version:**
Version 0.4.0

FabianTerhorst:
fix #646

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/PokemonGo.java
+++ b/library/src/main/java/com/pokegoapi/api/PokemonGo.java
@@ -132,6 +132,11 @@ public class PokemonGo {
 	 * @param altitude  the altitude
 	 */
 	public void setLocation(double latitude, double longitude, double altitude) {
+		if (latitude != this.latitude
+				|| longitude != this.longitude
+				|| altitude != this.altitude) {
+			getMap().clearCache();
+		}
 		setLatitude(latitude);
 		setLongitude(longitude);
 		setAltitude(altitude);
--- a/library/src/main/java/com/pokegoapi/api/map/Map.java
+++ b/library/src/main/java/com/pokegoapi/api/map/Map.java
@@ -689,6 +689,19 @@ public class Map {
 		return (api.currentTimeMillis() - lastMapUpdate) < api.getSettings().getMapSettings().getMinRefresh();
 	}
 
+	/**
+	 * Clear map objects cache
+	 *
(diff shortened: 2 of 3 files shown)


FabianTerhorst:
style warnings are fixed in #650

xae217:
Hey I just included this commit to my build but I'm gettinf a null pointer exception.

```
Exception in thread "Thread-0" java.lang.NullPointerException
    at com.pokegoapi.api.map.Map.clearCache(Map.java:697)
    at com.pokegoapi.api.PokemonGo.setLocation(PokemonGo.java:138)
    at poketest.PokeScanner.scan(PokeScanner.java:257)
    at poketest.PokeScanner.patrol(PokeScanner.java:331)
    at poketest.PokeScanner.run(PokeScanner.java:211)
    at java.lang.Thread.run(Thread.java:745)
```

Any ideas? For some reason cachedCatchable is null.

FabianTerhorst:
I will check for null then. I also don´t know when there are null.

FabianTerhorst:
there are setted to null instead of getting cleared

FabianTerhorst:
when the cache isn´t needed.

FabianTerhorst:
maybe we could clear the list instead of setting it null

FabianTerhorst:
@xae217 you can retest

xae217:
@FabianTerhorst It works now and seems to have fixed #646.

PR #654, opened 2016-08-22
remove network calls from api constructor

### LINKED ISSUE #610: New unused accounts error out on PokemonGo instance creation
**Description:**
I created a new PTC account and before even logging into the app on my phone tried to use the library to login.  This threw an error on creation of a PokemonGo instance.  I believe this is because until you accept the ToS, you can't use all the API calls (more specifically the settings calls).  Am trying to automated the tutorial process but this error causes problems

**Steps to reproduce:**
1. Create a PTC account
2. Create PokemonGo instance with new account

**Expected behavior:**
No errors thrown on PokemonGo instance creation.  Shouldn't do any API calls on instance creation except maybe login calls.

**Actual behavior:**
Error is thrown when Settings.updateSettings() is called from in the new PokemonGo(...) call

**Stacktrace (If it's a crash):**
http://pastebin.com/Jx6x7ufw

**Version:**
0.4.0 built from the Development branch

FabianTerhorst:
fix #610

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/PokemonGo.java
+++ b/library/src/main/java/com/pokegoapi/api/PokemonGo.java
@@ -47,7 +47,6 @@ public class PokemonGo {
 	private final byte[] sessionHash;
 	@Getter
 	RequestHandler requestHandler;
-	@Getter
 	private PlayerProfile playerProfile;
 	private Inventories inventories;
 	@Getter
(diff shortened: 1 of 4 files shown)


AbandonedCart:
Won't logging in inherit some sort of network exception. It seems impossible that an httpClient connection would throw no form of network error at all, so it made sense to have the constructor handle error without passing it along to something else.

Also, is there some justification for passing one throwable on to another throwable. With all the concern for efficiency, it seems like a wasted step to throw an error to another handler to throw it.

Paul776:
Thanks for this PR

FabianTerhorst:
will be closed in favor of #667

PR #659, opened 2016-08-23
Fix Signature checkstyle

AbandonedCart:
This separates fixing checkstyle from https://github.com/Grover-c13/PokeGOAPI-Java/pull/650 so they are no longer dependent upon each other, despite being somewhat unrelated.

### DIFF
--- a/library/src/main/java/com/pokegoapi/util/Signature.java
+++ b/library/src/main/java/com/pokegoapi/util/Signature.java
@@ -82,7 +82,8 @@ public class Signature {
 	}
 
 
-	private static int getLocationHash1(PokemonGo api, byte[] authTicket, RequestEnvelopeOuterClass.RequestEnvelope.Builder builder) {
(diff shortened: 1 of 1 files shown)


FabianTerhorst:
The builder is not needed in the method body.

FabianTerhorst:
There is also another method where the builder isn´t needed as well.

AbandonedCart:
Read the description. This was only intended to separate #650 and the checkstyle warning. 
Fixing checkstyle from a previous commit should not be dependent on also accepting all changes proposed in #650, but any proposed changes to the current code belong there.

AbandonedCart:
It should also be noted that the build failure is a false positive due to the branch being removed before the test was complete, but it has been proven that checkstyle is not documented by Travis anyway.

FabianTerhorst:
We should fix the check style output from travis.

PR #672, opened 2016-08-25
[WIP][Part3][Requres Part2 + Part 2.1]Rework login flow

FabianTerhorst:


FabianTerhorst:
We should thing about removing or deprecating the option to set a http client to the providers inside this pr.

LoungeKatt:
Is this complete in its current state? Disregarding additional plans for how to improve it, can this be deployed for the end user to begin reworking the flow in their code?

AbandonedCart:
This still needs a constructor that allows setting just the http client (since the only method to do that is by filling in every other value) unless that is completely removed and / or a check before declaring a new client in each of the other constructors otherwise it still declares a new one each time when not calling the one with all values.

FabianTerhorst:
We decided to pass the http client with the .login(provider) method.

PR #671, opened 2016-08-25
[WIP][Part1]Rework login flow

iGio90:
added locale to getPlayer

PR #673, opened 2016-08-25
Added some async method.

michibat01:


FabianTerhorst:
not the best timing, because of #671

michibat01:
how can I do? @FabianTerhorst

FabianTerhorst:
Could take some time until its getting merged. We will first merge dev in master

ramarro123:
asyncEnableAccount should be enableAccountAsync 
just to keep the naming convention

PR #680, opened 2016-08-26
[Part2.2]Rework login

FabianTerhorst:
Remove unneeded exceptions

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/gym/Battle.java
+++ b/library/src/main/java/com/pokegoapi/api/gym/Battle.java
@@ -15,6 +15,16 @@
 
 package com.pokegoapi.api.gym;
 
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.pokegoapi.api.PokemonGo;
+import com.pokegoapi.api.pokemon.Pokemon;
+import com.pokegoapi.exceptions.LoginFailedException;
+import com.pokegoapi.exceptions.RemoteServerException;
+import com.pokegoapi.main.ServerRequest;
+
+import java.util.ArrayList;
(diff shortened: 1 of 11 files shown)


library/src/main/java/com/pokegoapi/api/gym/Battle.java
```diff
@@ -126,7 +122,7 @@ public Result start() throws LoginFailedException, RemoteServerException {
 	 */
 	public AttackGymResponse attack(int times) throws LoginFailedException, RemoteServerException {
 
-		ArrayList<BattleAction> actions = new ArrayList<BattleAction>();
+		ArrayList<BattleAction> actions = new ArrayList<>();
```
AbandonedCart, line 125: There is no benefit to stripping the type declaration, but doing so could cause issues with some IDEs / compilers.

FabianTerhorst: Diamonds are supported since Java 7. That's the same Version gralde 3.0 needs to build.

AbandonedCart: Again, there is no benefit.

Jari27: There is no harm either, my IDE complains if I add redundant type arguments.

AbandonedCart: If it complains about providing a type definition, your IDE is broken.

Jari27: > redundant

PR #684, opened 2016-08-26
Added some async method.

michibat01:
Added updateProfileAsync
Added acceptLevelUpRewardsAsync
Added checkAndEquipBadgesAsync
Added enableAccountAsync
Added updateSettingsAsync


s7092910:
I'm not sure you understand the meaning of Async. While yes you are running the stuff on a different thread. You are blocking the thread that the call was made on to begin with.

PR #685, opened 2016-08-27
Rework client initialization

iGio90:
What does it do:
- Added constant class to hold our constants 
- Strip out the request from the constructor of the Settings and Inventories classes
- Improve the client initialization by firing the 2 block of calls that the official client is actually doing
- Store the hash of the Settings for later usage (see block 2 of login calls)

here is a gist with this impl:

https://gist.github.com/iGio90/180b53ee4e275e31dd15577c4165508c

This rework has been done together with @michibat01  👍

### DIFF
--- a/library/src/main/java/com/pokegoapi/util/Constant.java
+++ b/library/src/main/java/com/pokegoapi/util/Constant.java
@@ -5,4 +5,7 @@ package com.pokegoapi.util;
  */
 
 public class Constant {
+    public static final int APP_VERSION = 3500;
+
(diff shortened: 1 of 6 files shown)


library/src/main/java/com/pokegoapi/util/Constant.java
```diff
@@ -0,0 +1,11 @@
+package com.pokegoapi.util;
+
```
FabianTerhorst, line 2: Missing license header

PR #692, opened 2016-08-27
Get new accounts ready to use the library

iGio90:
The purpose of this commit, as for my previous (https://github.com/Grover-c13/PokeGOAPI-Java/pull/685) is to naturalize the way the library act to prevent the accounts to get flagged. 
<br />What i'm going to do in here, is to check if the current logged in account have an avatar, a nickname, and the first pokémon catched up. 
<br />Why? Very simple. The official client, before firing calls such as the getMapObject (the mostly common) is ALWAYS with a fully activated account (This means, with an avatar, a nickname, and the whole tutorial filled). There is no server side check on this, maybe because it's better for them to flag and ban instead of preventing (understandable for detecting bots and 3rd party clients). 
<br />So that, I'm going to force our library with a check that will setup a random avatar and nickname (probably based on the seed we are using to feed the device info and other things). 
<br />According to my dumps, the official client is ALWAYS firing multiple calls together, so at first, i reworked my old commit to set the tutorial state to the legal screen and append on the request some other commons requests that will update our classes as well.

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/PokemonGo.java
+++ b/library/src/main/java/com/pokegoapi/api/PokemonGo.java
@@ -17,6 +17,7 @@ package com.pokegoapi.api;
 
 import POGOProtos.Enums.PlatformOuterClass;
 import POGOProtos.Enums.PlatformOuterClass.Platform;
+import POGOProtos.Enums.TutorialStateOuterClass;
 import POGOProtos.Networking.Envelopes.RequestEnvelopeOuterClass.RequestEnvelope.AuthInfo;
 import POGOProtos.Networking.Envelopes.SignatureOuterClass;
 import POGOProtos.Networking.Requests.Messages.GetAssetDigestMessageOuterClass.GetAssetDigestMessage;
@@ -179,7 +180,8 @@ public class PokemonGo {
 
 		// We are going to set this a-side instead of inside the previous check for
 		// backward compatibility
-		if (!playerProfile.getAvatar().isAvatarCreated()) {
+		if (!playerProfile.getTutorialState().getTutorialStates().
+				contains(TutorialStateOuterClass.TutorialState.AVATAR_SELECTION)) {
 			playerProfile.initializeAccount();
 		}
 	}
--- a/library/src/main/java/com/pokegoapi/api/PokemonGo.java
+++ b/library/src/main/java/com/pokegoapi/api/PokemonGo.java
@@ -15,16 +15,12 @@
 
 package com.pokegoapi.api;
 
-import POGOProtos.Enums.PlatformOuterClass;
-import POGOProtos.Enums.PlatformOuterClass.Platform;
 import POGOProtos.Enums.TutorialStateOuterClass;
 import POGOProtos.Networking.Envelopes.RequestEnvelopeOuterClass.RequestEnvelope.AuthInfo;
 import POGOProtos.Networking.Envelopes.SignatureOuterClass;
-import POGOProtos.Networking.Requests.Messages.GetAssetDigestMessageOuterClass.GetAssetDigestMessage;
 import POGOProtos.Networking.Requests.RequestTypeOuterClass;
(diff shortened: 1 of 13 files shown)


iGio90:
last commit includes:
- stripped the 2 block of request to be re-used
- setup constants for avatar
- check if the avatar exist or create it (needs test)
- reword a bit the playerprofile class

iGio90:
Last chain of commits includes:
- Externalization of common request for re-use on demand
- EncounterTutorialComplete (A.K.A the catch of the first pokémon randomed between bulbasaur, charmender or squirtle)

iGio90:
Last 2 commits:
- Generify the array of common request
- ClaimCodeName tutorial step that will setup a random generate username between 10 and 15 char (max) + fail check and retry + mark the step of tutorial

iGio90:
Ok... I need to fix the checkstyle. Here are 2 dumps for comparison
Official: https://gist.github.com/iGio90/7522563dee535aadb71dc397cc11c6d8
Library: https://gist.github.com/iGio90/f7ce7378694a540f14cdb716355c36f4

GG? Lemme know
@FabianTerhorst @Grover-c13 @LoungeKatt

library/src/main/java/com/pokegoapi/api/PokemonGo.java
```diff
@@ -166,46 +160,70 @@ public void login(CredentialProvider credentialProvider) throws LoginFailedExcep
...
+
+		fireRequestBlockTwo();
+
+		// From now one we will start to check our accounts is ready to fire requests.
+		// Actually, we can receive valid responses even with this first check,
+		// that mark the tutorial state into LEGAL_SCREEN.
+		// Following, we are going to check if the account binded to this session
+		// have an avatar, a nickname, and all the other things that are usually filled
+		// on the official client BEFORE sending any requests such as the getMapObject etc.
+		if (playerProfile.getTutorialState().getTutorialStates().isEmpty()) {
```
FabianTerhorst: Can you assign it to a Value and Reuse this Value for the other Checks?

FabianTerhorst: PlayerProfile.getTutorialState().getTutorialStates()

iGio90:
What you mean?

iGio90:
Salck pm with explain if you can!

library/src/main/java/com/pokegoapi/api/PokemonGo.java
iGio90: yeah! i can :) gonna do it within checkstyle fixes

iGio90: yeah! i can :) gonna do it within checkstyle fixes

Paul776:
@iGio90 I have compared both gist. Looks good to me. 👍

library/src/main/java/com/pokegoapi/api/PokemonGo.java
```diff
@@ -166,46 +161,68 @@ public void login(CredentialProvider credentialProvider) throws LoginFailedExcep
...
+			playerProfile.firstTimeExperienceComplete();
+		}
+	}
+
+	/**
+	 * First requests block. Private since we will use this only at initialization!
+	 *
+	 * @throws LoginFailedException  When login fails
+	 * @throws RemoteServerException When server fails
+     */
```
FabianTerhorst: Doc end is wrong alligned. As well as some others. Just indent the file.

PokeAlert:
This is a dump of this PR

```
PokemonGo pokemon = new PokemonGo(Utils.http, account.getSeed());
pokemon.login(credentialProvider);
```

result in:
https://gist.github.com/PokeAlert/2236ab17282580d39d17f7f451024cdc

original client (0.33) from login to the end of the tuto
https://gist.github.com/PokeAlert/3ebd8c4a41eea4b866e9debc22006070

original client (0.35) from login to end of tuto
https://gist.github.com/PokeAlert/ee1c80a79aa68089e0febb4ec23c3122

PR #695, opened 2016-08-27
Fix clearing the cache and refilling lists

AbandonedCart:
- Do not clear the cache to move to the exact spot at higher altitude
- Check that both the cache is to be used AND there are objects in it

Note: There is no master check for cached map objects, so the most
commonly filled object was used instead (cachedCatchable)


AbandonedCart:
Preemptive explanation:
The inline check against useCache will handle verifying use to return the cached object and if it should continue beyond that point, then https://github.com/AbandonedCart/PokeGOAPI-Java/blob/910078b8e09528b38bae65e17b4304f0be7e54ed/library/src/main/java/com/pokegoapi/api/map/Map.java#L120 will clear the cache before new objects are injected (making it redundant)

PR #704, opened 2016-08-28
Map: append common requests to GetMapObject

iGio90:
I'll wait for your ok to merge this but imho, is good to have this inside.
getMapObject request on official client is always append on top of the other 4 common requests. 
From what we can understand from the dump the logic of how the requests are fired is quite simple. There is a loop of 5 requests that are always fired together. The first request, on top of the other, could be anything (such as the getMapObject or others like DownloadItemTemplates etc.etc.). All of these, are followed by a set of 4 requests that are now inside the CommonRequest class. 

O.T: I'll provide later on a more generalization on how the responses are managed by the library. Actually I'm copy pasting how the responses of the 4 subsequent calls are handled, but as said, I'm going to generalize also these.

### DIFF
--- a/library/src/main/java/com/pokegoapi/main/CommonRequest.java
+++ b/library/src/main/java/com/pokegoapi/main/CommonRequest.java
@@ -104,4 +104,24 @@ public class CommonRequest {
 				CommonRequest.getDownloadSettingsMessageRequest(api));
 		return serverRequests;
 	}
+
+	/**
+	 * List of common requests
+	 *
+	 * @param api The current instance of PokemonGO
+	 * @return an array of ServerRequest
+	 *
+	 */
+	public static ServerRequest[] commonRequests(PokemonGo api) {
+		ServerRequest[] serverRequests = new ServerRequest[4];
+		serverRequests[0] = new ServerRequest(RequestType.GET_HATCHED_EGGS,
+				GetHatchedEggsMessage.getDefaultInstance());
+		serverRequests[1] = new ServerRequest(RequestType.GET_INVENTORY,
+				CommonRequest.getDefaultGetInventoryMessage(api));
+		serverRequests[2] = new ServerRequest(RequestType.CHECK_AWARDED_BADGES,
+				CheckAwardedBadgesMessage.getDefaultInstance());
+		serverRequests[3] = new ServerRequest(RequestType.DOWNLOAD_SETTINGS,
+				CommonRequest.getDownloadSettingsMessageRequest(api));
+		return serverRequests;
+	}
 }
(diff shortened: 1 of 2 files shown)


ramarro123:
i don't think that this guarantee that the requests are correctly bundled.
workQueue (on RequestHandler) can be drained at any moment.

so it could happend that you send MapObject req followed (into different reqs) by the commonrequests and not bundled as you expect.

you should add some sort of lock on RequestHandler i think to make sure that the request are bundled as expected (you can do a lock before preparing the map request, and unlock after bundling all the requests)

i know that the 350ms window it's limited, but it will be still easy to spot this pattern from niantic if (and they probably do) log every single call

iGio90:
mhh or maybe swap from asyncserverreqesy to serverrequest?

ramarro123:
i think that's right.

but, the main problem for me, it's that the common-requests aren't added just on getMapObject
but on almost every requests that i see going around on MITM (battles, catch pokemon, etc etc)

i think that there should be a more "generic" way to add it.
what about create a new 
AsyncServerRequest (or add a custom boolean to it)
like "addCommonRequest /true/false"

and inject the commonRequest directly on the run() of RequestHandler?

so, in any method where you want to inject it you'll do something like

```
final AsyncServerRequest asyncServerRequest = new AsyncServerRequest(
                RequestType.GET_MAP_OBJECTS, builder.build(), true);

```

where the last parameters say "inject common requests"

iGio90:
Its a good idea! We have tons of way to do it! You can make a PR with your idea or you wanna wait me get up from the bad? :)

ramarro123:
i am not very good on git to make a pr, i have try in the past but it took 1 min to write the code and 2 days to upload it :(

iGio90:
No problem! Its very easy! I can do this when I get awake and i can also teach you how to do a pr in 30 sec

ramarro123:
ok, i'll start to write the code and ask for your assistance :)
in case will you mind to close this PR if i submit another one?

iGio90:
Of course i'll! If you then need assistence with the PR, reach me on slack and we are going to make it :)

library/src/main/java/com/pokegoapi/main/CommonRequest.java
```diff
@@ -104,4 +104,24 @@ public static GetInventoryMessage getDefaultGetInventoryMessage(PokemonGo api) {
...
+
+	/**
+	 * List of common requests
+	 *
+	 * @param api The current instance of PokemonGO
+	 * @return an array of ServerRequest
+	 *
+	 */
+	public static ServerRequest[] commonRequests(PokemonGo api) {
+		ServerRequest[] serverRequests = new ServerRequest[4];
```
FabianTerhorst, line 116: you can also init the array with the values = { 
new ServerRequest(RequestType.GET_HATCHED_EGGS,
-              GetHatchedEggsMessage.getDefaultInstance()),....
  }

ramarro123:
ok, i have do a pr request, if you want have a look and let me know if you like it
it's #711

at the moment i can't test it, seems that the lib it's broken somewhere during the login

PR #707, opened 2016-08-29
[DO NOT MERGE]Add check challenge in the common requests

PokeAlert:
client 0.35 sends the checkChallenge request along with the other common requests.

### DIFF
--- a/library/src/main/java/com/pokegoapi/main/CommonRequest.java
+++ b/library/src/main/java/com/pokegoapi/main/CommonRequest.java
@@ -15,13 +15,13 @@
 
 package com.pokegoapi.main;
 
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/pokegoapi/main/CommonRequest.java
```diff
@@ -15,6 +15,7 @@
 
 package com.pokegoapi.main;
 
+import POGOProtos.Networking.Requests.Messages.CheckChallenge;
```
AbandonedCart: This should be moved to line 24

nanowen:
Why the IV filter was erased? Now the app notifies me when there are low IV Pokémon and it is annoying.

PR #711, opened 2016-08-29
[WIP] add bundling request for common info

ramarro123:
on AsyncServerRequest
it's now possible to add an optional constructor (boolean)

when AsyncServerRequest it's created with autoBundle param
he will add all the method on 
commonRequests
in the same calls.

doesnt' work right now, but the current for doesnt' work, it return
com.pokegoapi.exceptions.AsyncPokemonGoException: Unknown exception occurred

even without this patch appied :(

### DIFF
--- a/library/src/main/java/com/pokegoapi/main/CommonRequest.java
+++ b/library/src/main/java/com/pokegoapi/main/CommonRequest.java
@@ -22,6 +22,7 @@ import POGOProtos.Networking.Requests.Messages.DownloadSettingsMessageOuterClass
 import POGOProtos.Networking.Requests.Messages.GetAssetDigestMessageOuterClass.GetAssetDigestMessage;
 import POGOProtos.Networking.Requests.Messages.GetHatchedEggsMessageOuterClass.GetHatchedEggsMessage;
 import POGOProtos.Networking.Requests.Messages.GetInventoryMessageOuterClass.GetInventoryMessage;
+import POGOProtos.Networking.Requests.RequestTypeOuterClass;
 import POGOProtos.Networking.Requests.RequestTypeOuterClass.RequestType;
 import POGOProtos.Networking.Responses.DownloadSettingsResponseOuterClass;
 import POGOProtos.Networking.Responses.GetInventoryResponseOuterClass;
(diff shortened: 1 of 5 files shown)


library/src/main/java/com/pokegoapi/main/CommonRequest.java
```diff
@@ -104,4 +108,50 @@ public static GetInventoryMessage getDefaultGetInventoryMessage(PokemonGo api) {
...
+
+
+	/**
+	 * List of common requests
+	 *
+	 * @param api The current instance of PokemonGO
+	 * @return an array of ServerRequest
+	 */
+	public static ServerRequest[] commonRequests(PokemonGo api) {
+		ServerRequest[] serverRequests = new ServerRequest[4];
```
FabianTerhorst: you can init the array with values instead of init and then set

ramarro123: i prefer to leave it as it is if you don't mind.
In the future can be possible that common requests will be  much more than 4.
i was thinking of adding something like

```
final int GET_HATCHED_EGGS_CR = 0;
final int GET_INVENTORY_CR = 1;
final int CHECK_AWARDED_BADGES_CR = 2;
```

and so on... 
this will make the parse method easy to read and implement with something like

```
 switch()
 case GET_HATCHED_EGGS_CR:
 case GET_INVENTORY_CR:
 case CHECK_AWARDED_BADGES_CR:
```

the compiler will inline them in any case

FabianTerhorst: RequestType.GET_HATCHED_EGGS is already an enum so you don´t have to add extra constants

library/src/main/java/com/pokegoapi/main/CommonRequest.java
```diff
@@ -104,4 +108,50 @@ public static GetInventoryMessage getDefaultGetInventoryMessage(PokemonGo api) {
...
+	}
+
+	/**
+	 * parse the response received during commonRequest
+	 *
+	 * @param api The current instance of PokemonGO
+	 * @param id The id of the current common request
+	 * @param data The data received from server
+	 */
+	public static void parse(PokemonGo api, int id, ByteString data) {
```
iGio90: Can we use the requestType instead of the "int id" here?

library/src/main/java/com/pokegoapi/main/CommonRequest.java
ramarro123: yes... but if we use the existent ebum we cant use them to return an array, ordinal value arent what will be used..on request handler its important to maintain the order of requests. we can convert into an enum map to make it even less readable if you want :)

ramarro123:
sure ill rename it later

iGio90:
@ramarro123 can you reach me on slack?

ramarro123:
yep how? i am not familiar with slack :)

ramarro123:
yep how? i am not familiar with slack :)

iGio90:
there is the link on the bottom of the readme, then join the channel #javaapi and write something like @iGio90 hey!

PR #714, opened 2016-08-29
update protos and remove sensorinfo, deviceinfo and activity status setters in favor of getBuilder

FabianTerhorst:


library/src/main/java/com/pokegoapi/api/device/LocationFixes.java
```diff
@@ -107,7 +107,11 @@ public static LocationFixes getDefault(PokemonGo api, RequestEnvelopeOuterClass.
 					.setAltitude(altitude)
 					.setVerticalAccuracy(verticalAccuracy)
 					.setProviderStatus(3)
-					.setLocationType(1);
+					.setLocationType(1)
+					.setAltitude(api.getAltitude())
+					.setSpeed(/*Todo : speed iOS only (-1 for no reading available, speed in m/s)*/0)
+					.setCourse(/*Todo : iOS only (range seems to be -1 for not reading, and 0 to 360 for reading*/0)
+					.setFloor(/*Todo : The floor of the building this person is on*/0);
```
FabianTerhorst, line 114: @iGio90 you want to fill the 3 new values as well?

iGio90: i was already looking at it

FabianTerhorst: you found something?

FabianTerhorst: setCourse is just a random between 0 and 360

Grover-c13:
course & speed can be -1 as well, still valid.

FabianTerhorst:
ok

FabianTerhorst:
and floor? ^^

iGio90:
yeah, but is still better to findout a valid logic to random a good value in a good range. I'll take it please! :)

FabianTerhorst:
i can just create a random for 0-360

iGio90:
I would like to check before if it's really a random between 0-360 or there is a logic that could say if (accel_x > X || accel_y < Y) (or similar) course = Z

FabianTerhorst:
normally the course shouldn´t have any subjections from other values

FabianTerhorst:
for the speed we just have to calculate the distance between the last lat and lng and next and adding them to a value. This value is for the distance the person travelled since app start and calc that down to sec.

library/src/main/java/com/pokegoapi/api/PokemonGo.java
```diff
@@ -273,11 +294,14 @@ public AuthInfo getAuthInfo()
 	 *
 	 * @param latitude  the latitude
 	 * @param longitude the longitude
-	 * @param altitude the altitude
+	 * @param altitude  the altitude
 	 * @param accuracy  the accuracy
 	 */
 	public void setLocation(double latitude, double longitude, float altitude, double accuracy) {
 		if (latitude != this.latitude || longitude != this.longitude) {
+			if (this.latitude != Double.NaN && this.longitude != Double.NaN) {
```
AbandonedCart: The default for each is NaN, so unless someone explicitly sets one to "Double.NaN", any time the previous check is met, this check will always be true.

AbandonedCart:
This is inconsistent with device info and activity status, plus it moves toward preventing the user from setting their own values. While many will opt for defaults, removing the ability to easily set them as the end user is a huge regression.

FabianTerhorst:
you can still set the own properties. Just sensorInfo.getBuilder()

FabianTerhorst:
we should do the same for activitystatus and deviceinfo

FabianTerhorst:
Now its not inconsistent anymore

FabianTerhorst:
@iGio90 whats your opinion?

FabianTerhorst:
should i make floor = altitude?

library/src/main/java/com/pokegoapi/api/PokemonGo.java
```diff
@@ -311,7 +366,7 @@ public void setLongitude(double value) {
 	 * @throws IllegalStateException if location has not been set
 	 */
 	public Map getMap() {
-		if (this.latitude == Double.NaN || this.longitude == Double.NaN) {
```
s7092910, line 314: Should keep this. As there is still a single setter for longitude. Unless you plan to set the single setters for any location related value to private.

PR #717, opened 2016-08-29
CommonRequest: Added CheckChallenge and append it to the twice login …

iGio90:


library/src/main/java/com/pokegoapi/main/CommonRequest.java
```diff
@@ -83,6 +84,20 @@ public static GetInventoryMessage getDefaultGetInventoryMessage(PokemonGo api) {
 	}
 
 	/**
+	 * Append CheckChallenge request to the given ServerRequest
+	 *
+	 * @param request The main request we want to fire
+	 * @return an array of ServerRequest
+     */
+	public static ServerRequest[] appendCheckChallenge(ServerRequest request) {
+		ServerRequest[] serverRequests = new ServerRequest[2];
```
FabianTerhorst: init with values?

PR #720, opened 2016-08-29
Prefer user values to maintain consistency

AbandonedCart:
A random should not be used when a user value exists, since this is supposed to be signing a call by wrapping it with what should be an extended reading of the same device in the same place.

SensorInfo should favor user values before attempting to instantiate / update a default

ActivityStatus should allow user configuration and favor it before instantiating a default

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/PokemonGo.java
+++ b/library/src/main/java/com/pokegoapi/api/PokemonGo.java
@@ -328,4 +328,18 @@ public class PokemonGo {
 		}
 		return deviceInfo.getDeviceInfo();
 	}
+	
+	/**
+	 * Gets the sensor info
+	 *
+	 * @param currentTime the current time
+	 * @param random      the random object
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/pokegoapi/api/PokemonGo.java
```diff
@@ -327,4 +328,31 @@ public Map getMap() {
...
+	}
+	
+	/**
+	 * Gets the activity status
+	 *
+	 * @param random the random object
+	 * @return the activity status
+	 */
+	public SignatureOuterClass.Signature.ActivityStatus getActivitySignature(Random random) {
+		if (activityStatus.getActivityStatus() == null) {
```
FabianTerhorst, line 353: This is not null save

FabianTerhorst: Check activity Status for null. Not the attribute.

PR #721, opened 2016-08-29
Rework request handler and AsyncServer requests

iGio90:
Reworked to get bounded with common requests


iGio90:
Thanks to @ramarro123 for some logical code!

iGio90:
@FabianTerhorst @LoungeKatt this is also ready to be merged

PR #728, opened 2016-08-30
Update PokemonGo.java

PokeBotPub:
Added setDeviceInfo


AbandonedCart:
This is already done when it is called at https://github.com/PokeBotPub/PokeGOAPI-Java/blob/581e9081bb6e445911d0c161559fcb54b8195aa5/library/src/main/java/com/pokegoapi/api/PokemonGo.java#L336 by setting the device info through the DeviceInfo class.

PR #729, opened 2016-08-30
[WIP - WIP - WIP] Rework everything

iGio90:


library/src/main/java/com/pokegoapi/api/pokemon/Pokemon.java
```diff
@@ -246,7 +246,7 @@ public boolean canEvolve() {
 	public EvolutionResult evolve() throws LoginFailedException, RemoteServerException {
 		EvolvePokemonMessage reqMsg = EvolvePokemonMessage.newBuilder().setPokemonId(getId()).build();
 
-		Pokemon me = this;
+		final Pokemon me = this;
```
FabianTerhorst: are you sure the reference isn´t strong enough?

FabianTerhorst: I don´t think this value is even needed.

library/src/main/java/com/pokegoapi/api/inventory/ItemBag.java
```diff
@@ -72,7 +72,7 @@ public void addItem(Item item) {
 	 * @throws LoginFailedException  the login failed exception
 	 */
 	public Result removeItem(ItemId id, int quantity) throws RemoteServerException, LoginFailedException {
-		Item item = getItem(id);
+		final Item item = getItem(id);
```
FabianTerhorst, line 73: Is this not out of scope for the pr?

library/src/main/java/com/pokegoapi/main/RequestHandler.java
```diff
@@ -274,65 +274,58 @@ private Long getRequestId() {
...
-				for (AsyncServerRequest request : requests) {
-					resultMap.put(request.getId(), ResultOrException.getError(e));
-				}
+				resultMap.put(request.getId(), ResultOrException.getError(e));
 				continue;
 			} finally {
-				requests.clear();
+				try {
+					Thread.sleep(500);
+				} catch (Throwable e) {
```
FabianTerhorst: can you just name the full exception that can be thrown?

FabianTerhorst: InterruptedException

library/src/main/java/com/pokegoapi/api/inventory/ItemBag.java
ramarro123: no, the scope it's to remove all the sync requests and remove the method on requesthandler

library/src/main/java/com/pokegoapi/api/map/Map.java
```diff
@@ -584,90 +586,30 @@ public FortSearchResponse searchFort(FortData fortData) throws LoginFailedExcept
...
-	 * @param normalizedHitPosition the normalized hit position
-	 * @param normalizedReticleSize the normalized reticle size
-	 * @param spinModifier          the spin modifier
-	 * @param pokeball              the pokeball
-	 * @return the catch pokemon response
-	 * @throws LoginFailedException  the login failed exception
-	 * @throws RemoteServerException the remote server exception
-	 */
-	@Deprecated
-	public CatchPokemonResponse catchPokemon(
```
FabianTerhorst, line 643: you removed 2 deprecated methods. Is this still in scope?

ramarro123: yes, they use synchronous method

ramarro123: synchronous method = bye bye :)
methods that are handled on commonRequest = bye bye

also, all the lists that will throw modification exceptions have to go, same for items that have duplicate stuff on them, have to be converted on map. clear method on inventory will go and inventory need to be updated in a proper way, not cleaning and populating it every time

FabianTerhorst:
Looks good right now when you rename the Throwable to a InterruptedException and when you just remove the Pokemon me = this; reference.

ramarro123:
i htink that pokemon me = this it's necessary for the remove

FabianTerhorst:
and you have to fix the sample.

ramarro123:
sure, it's still a wip, there is still a lot of work to do in the library, we'll do the sample as well

iGio90:
@FabianTerhorst is not necessary to analyze the code right now. The work is still a lot. I.E actually since we are moved all the calls to Async, we will need to code a way to provide errors and throws. We will fix sample and all the javadocs as well :)

ramarro123:
on battle & gym i will comment all the function that use sync calls.

i was never able to make them work, and i think that the code it's very buggy (i got flagged and banned after 1 try) so, i suggest to leave it commented unless you want to take care of it (using mitm and such)

pvanassen:
You could take a look at https://github.com/pvanassen/PokeGOAPI-Java/tree/mimic for inspiration. It's a fork of this API, everything is async, no blocking methods. 

Won't merge that code because it serves a different goal, imitating the app as good as possible.

iGio90:
We won't take inspiration! We are doing it in our way :D anyway, we will also try to imitate everything that the official client do.

iGio90:
Buuut, i'll give it a look once im back home. We are also getting rid of all the sync blocking calls

pvanassen:
Are you going to make the Map update itself every 10 seconds or are you going to do this on request? I did the first making the Map API a lot easier to use.

iGio90:
I'm thinking on a settable heartbeat. Requests are already bounded with the 6 common on the same envelope, so it would be best to make an heartbeat to full spoof the client

pvanassen:
That way you can make most map calls without Observable, or provide an Observable which never calls OnComplete, but only OnNext's with new pokemon

pvanassen:
Let me know if you need a hand. Happy to chip in.

iGio90:
You can pm me on slack if you want :)

ramarro123:
so, guys, i don't know if you will be intrested on this PR, i have changed quite a lot...

1) remove all sync method
2) removed RX in favor of a simpler interface method, to mimic client we need to send request one by one (bundled or not) with a throttle, RX pattern was not that usefull, so i removed the dependency
3) all the game method will require a callback, and are async of course.
4) the request handler it's now begin rewritten to add the commoninfo and parse them.
5) map object will be restyled in the next few days.

at the moment the code of course doesn't work.

when we will finish, will be great if someone can check all the methods to ensure that we haven't forget something on the new parsing function.

Also, due to the fact that we are moving to a full async pattern, all the blocking utils will be deleted, all the deprecated method will be deleted, gym & battle will be deleted cos it's not maintained afaik (feel free to join us if you do)

A lot of data structure will be replaced with concurrentMap or synchronized block or things like that, to make it threadsafe.

forgot to add:
parsing & callbacks are optionals, so for method that doesn't need to return or do anything, there is nothing to do.
callbacks support error notification, with a default log (can be overridden of course)

library/src/main/java/com/pokegoapi/api/inventory/CandyJar.java
```diff
@@ -15,25 +15,21 @@
...
+import com.pokegoapi.api.PokemonGo;
 import lombok.ToString;
 
+import java.util.concurrent.ConcurrentHashMap;
+
 @ToString
 public class CandyJar {
 	private final PokemonGo api;
-	private final HashMap<PokemonFamilyId, Integer> candies = new HashMap<>();
+	private final ConcurrentHashMap<PokemonFamilyId, Integer> candies = new ConcurrentHashMap<>();
```
PokeAlert: Use the interface ConcurrentMap

library/src/main/java/com/pokegoapi/api/inventory/CandyJar.java
```diff
@@ -66,14 +50,16 @@ public void addCandy(PokemonFamilyId family, int amount) {
 	 * @param amount Amount of candies to remove
 	 */
 	public void removeCandy(PokemonFamilyId family, int amount) {
-		if (candies.containsKey(family)) {
-			if (candies.get(family) - amount < 0) {
-				candies.put(family, 0);
+		synchronized (candies) {
```
PokeAlert: don't use synchronized use putIfAsbent and check return value

library/src/main/java/com/pokegoapi/api/inventory/Hatchery.java
```diff
@@ -15,70 +15,66 @@
...
-import POGOProtos.Networking.Requests.RequestTypeOuterClass.RequestType;
-import POGOProtos.Networking.Responses.GetHatchedEggsResponseOuterClass.GetHatchedEggsResponse;
-import lombok.Getter;
+import java.util.concurrent.ConcurrentHashMap;
 
 public class Hatchery {
 	@Getter
-	private final Set<EggPokemon> eggs = new HashSet<EggPokemon>();
+
+	private final ConcurrentHashMap<Long, EggPokemon> eggs = new ConcurrentHashMap<>();
```
PokeAlert: Use ConcurrentMap interface

library/src/main/java/com/pokegoapi/api/inventory/Hatchery.java
```diff
@@ -15,70 +15,66 @@
...
 
-	public void reset() {
-		eggs.clear();
-	}
-
-	public void addEgg(EggPokemon egg) {
-		egg.setApi(api);
-		eggs.add(egg);
+	public void addEgg(PokemonDataOuterClass.PokemonData egg) {
+		synchronized (eggs) {
```
PokeAlert: remove synchcronised use putIfAbsent and check returned value

library/src/main/java/com/pokegoapi/api/inventory/Inventories.java
```diff
@@ -52,7 +41,7 @@
 	@Getter
 	private Pokedex pokedex;
 	@Getter
-	private final List<EggIncubator> incubators = new ArrayList<>();
+	private final ConcurrentHashMap<String, EggIncubator> incubators = new ConcurrentHashMap<>();
```
PokeAlert: ConcurrentMap

library/src/main/java/com/pokegoapi/api/inventory/Inventories.java
```diff
@@ -162,11 +107,21 @@ public void updateInventories(GetInventoryResponse response) {
 
 			if (itemData.hasEggIncubators()) {
 				for (EggIncubatorOuterClass.EggIncubator incubator : itemData.getEggIncubators().getEggIncubatorList()) {
-					incubators.add(new EggIncubator(api, incubator));
+					synchronized (incubators) {
```
PokeAlert: remove synch see other comments

library/src/main/java/com/pokegoapi/api/inventory/ItemBag.java
```diff
@@ -15,81 +15,75 @@
...
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * The type Bag.
  */
 public class ItemBag {
 	private final PokemonGo api;
-	private final HashMap<ItemId, Item> items = new HashMap<>();
+	private final ConcurrentHashMap<ItemId, Item> items = new ConcurrentHashMap<>();
```
PokeAlert: ConcurrentMap

library/src/main/java/com/pokegoapi/api/inventory/ItemBag.java
```diff
@@ -15,81 +15,75 @@
...
 	}
 
-	public void reset() {
-		items.clear();
-	}
-
-	public void addItem(Item item) {
-		items.put(item.getItemId(), item);
+	public void addItem(ItemData item) {
+		synchronized (items) {
```
PokeAlert: remove synch, see other comments

library/src/main/java/com/pokegoapi/api/inventory/PokeBank.java
```diff
@@ -15,44 +15,38 @@
...
-
-import POGOProtos.Enums.PokemonIdOuterClass;
-import lombok.Getter;
+import java.util.concurrent.ConcurrentHashMap;
 
 
 public class PokeBank {
 	@Getter
-	private final List<Pokemon> pokemons = Collections.synchronizedList(new ArrayList<Pokemon>());
+	private final ConcurrentHashMap<Long, Pokemon> pokemon = new ConcurrentHashMap<Long, Pokemon>();
```
PokeAlert: ConcurrentMap

library/src/main/java/com/pokegoapi/api/inventory/PokeBank.java
```diff
@@ -15,44 +15,38 @@
...
 	 *
 	 * @param pokemon Pokemon to add to the inventory
 	 */
-	public void addPokemon(final Pokemon pokemon) {
-		List<Pokemon> alreadyAdded = Stream.of(pokemons).filter(new Predicate<Pokemon>() {
-			@Override
-			public boolean test(Pokemon testPokemon) {
-				return pokemon.getId() == testPokemon.getId();
+	public void addPokemon(PokemonGo api, PokemonDataOuterClass.PokemonData pokemonData) {
+		synchronized (pokemon) {
```
PokeAlert: remove synch, see other comments

library/src/main/java/com/pokegoapi/api/inventory/PokeBank.java
```diff
@@ -63,12 +57,12 @@ public boolean test(Pokemon testPokemon) {
...
 	public List<Pokemon> getPokemonByPokemonId(final PokemonIdOuterClass.PokemonId id) {
-		return Stream.of(pokemons).filter(new Predicate<Pokemon>() {
-			@Override
-			public boolean test(Pokemon pokemon) {
-				return pokemon.getPokemonId().equals(id);
-			}
-		}).collect(Collectors.<Pokemon>toList());
+		List<Pokemon> ret = new ArrayList<>();
+		for (Pokemon p : pokemon.values()) {
+			if (p.getPokemonId() == id)
```
PokeAlert: {
}

library/src/main/java/com/pokegoapi/api/map/Map.java
```diff
@@ -83,410 +53,66 @@
...
-							java.util.Map<FortType, List<FortData>> groupedForts = Stream.of(mapCell.getFortsList())
-									.collect(Collectors.groupingBy(new Function<FortData, FortType>() {
-										@Override
-										public FortType apply(FortData fortData) {
-											return fortData.getType();
-										}
-									}));
-							result.addGyms(groupedForts.get(FortType.GYM));
-							result.addPokestops(groupedForts.get(FortType.CHECKPOINT));
+							HashSet<FortDataOuterClass.FortData> gyms = new HashSet<>();
```
PokeAlert: Set

library/src/main/java/com/pokegoapi/api/map/Map.java
```diff
@@ -83,410 +53,66 @@
...
+							HashSet<FortDataOuterClass.FortData> pokestops = new HashSet<>();
```
PokeAlert: Set

library/src/main/java/com/pokegoapi/api/inventory/CandyJar.java
ramarro123: i don't think that you can, it's a delta value. you haven't to put something you have to get the current val, subtract and put.
you can use compute but it's jdk8 only

library/src/main/java/com/pokegoapi/api/inventory/Hatchery.java
ramarro123: any particular reason for doing that?

ramarro123:
ok guys, take a look to the code and let us know what do you think of it

forgot to add:
the last step will be add an heartbeat to the map object, but first i will collect all the feedback for the work done so far, and add it later

Paul776:
hmm....
Can we PLEASE have an synch option? having this call back thing make it VERY VERY hard to work with threads and executors.

A typical worker looks like this

```
public void run() {
   try {    
         List<..> pokemons = api.getCatchable(..)
     } catch (...) {
     } finally {
             bus.sendmessage(clear scan circle)
             whatever..
     }
}
```

And now with the callback?? impossible without making super complex. or putting call inside the onResponse

```
> >  api.getPoke(new Listener() {
> >             onResponse() {
> >             }
> >             onFail() {
> >                 // I want to retry when this happens how the hell do i do now? 
> >              }
> >      }    
> 
```

iGio90:
The onresponse provide exactly the objects you are looking for. Maybe try it before? :D

iGio90:
And about the error, yeah, it would be simple to add a RequestType to understand which call throw the error

Paul776:
Having the object that's not the problem. The problem is that with the getMap() being asynch I lose the ability to control the flow of my thread. the finally block will be called before (maybe) the onResponse/onfail gets called 

it makes it super hard to handle errors correctly. Before when it was synch we could have different catch. For example when getCatchablePokemon was throwing a LoginException I was catching that and automatically re-login with retries and then retry the getCatchablePokemon

and now? super hard with .login being synch swell

```
api.getMap()...{
    public void onError(Throwable throwable) {
                if throwable instance of.. LoginExcpetion { 
                     // and now what?? My original thread is dead already and each thread respect a 10 sc   interval between poke requests 
                     //I want to login?
                     api.login() ...
                       and then on onResponse what ? it goes to deep
                }
            }
}
```

iGio90:
Umh... I dont know how you are handling things inside your app. I reworked half of my app yesterday and everything works fine. You can keep your threads alive in many ways (specially in android) o.O

ramarro123:
its super simple...in your main loop do a if (logged)  and set logged = false on exception. the main loop will relogin it.

its necessary to move on async imho to guaranter that niantix dont ban you

Paul776:
.login is asych.... you cannot do this anymore. how to wait .login returns something?

You can keep your inner internal stuff asynch, no problem with that. but you should provide a way to get the final result non-asynch (as well as asynch so the api user has the choice)

Paul776:
```
api.getMap(..., 
         onResponse(Map map){}
         onFail(Exception e) {
               // 1st problem there is no way to know whats happend here
               if (e instance of LoginException) {
                      api.login(..., 
                        onResponse(Map map){
                               api.getMap (....)
                                // etc good luck with that you cannot even do it recursively because don't forget I'm inside a Thread's run() and once you arrive here your thread has died already
                         }
                        onFail(Exception e) {
                             // got forbid me if it fails for any reason I wish to try again.. I need to go an other level down
                         }
         }
```

To complex for something that can be done really easily if getMap returns something and throws something

ramarro123:
you can made your own callback class and make it wait till the call it's over

class callback extend pokecallback() {
error;
response;
onerror() {done = true}
onresponse(){done = true}
await(){while (!done){sleep(1)}}
callbac() {
done = false;
}
}

and use this class as
callback c = new callback();
login(c)
c.await();
c.getError();
c.getResponse();

Paul776:
Please tell me you didn't put this kind of code in the api 
"{while (!done){sleep(1)}}"

You have multiple way of writing a 'workaround' (CountDownLatch for example) but the code you suggest must be the worst solution

ramarro123:
cos i think that blocking its not good at all :) and i propose you the quickeat solution. you can do the blocking pr uf you want after this one get merged eventually. i will not work on that, blocking it's something that i have removed, i dont like to add it back at all....

and sorry for the super lazy example was just a proof of concept. review the code on this pr to check if there are any issue that you want to share with us (apart of course the fact that its async which its intended)

iGio90:
Im not understanding at all what's your problem with the async way lol... you can handle retries inside the callback like

```
                        if (e.getMessage().contains("banned")) {
                            activity.runOnUiThread(new Runnable() {
                                @Override
                                public void run() {
                                    if (activity instanceof LoginActivity) {
                                        ((LoginActivity) activity).onAsyncException();
                                    }
                                }
                            });
                        } else {
                            activity.runOnUiThread(new Runnable() {
                                @Override
                                public void run() {
                                    if (activity instanceof LoginActivity) {
                                        ((LoginActivity) activity).onLoginFailed();
                                    }
                                }
                            });
                        }
```

O.o

pvanassen:
No observables anymore, that's a real blocker to me.

ramarro123:
so you prefer that method return an observable and add an observer to it?

something like doRequest(). addObserver(new Observer(){})
instead of
doRequest(new Callback(){})

it's not that different, the same code that you have to write on the
update method, it's what you have to write on the callback or i miss something?

pvanassen:
It's always easy to find some example that exactly fits _your_ argument why you love Callbacks. 

I like Observables because it gives me cleaner code. RxJava handles a lot of the complicated stuff around threading and locking. I don't like callbacks because it is very very hard to do right. Things like "{while (!done){sleep(1)}}" are a very clear indication things are not done right.

library/src/main/java/com/pokegoapi/main/RequestHandler.java
```diff
@@ -274,68 +198,67 @@ private Long getRequestId() {
...
-					resultMap.put(request.getId(), ResultOrException.getError(e));
-				}
+				request.fire(e);
 				continue;
 			} finally {
-				requests.clear();
+				try {
+					Thread.sleep(350);
+				} catch (InterruptedException ignored) {
+					// Ignore
```
pvanassen: Ignoring an InterruptedException is not very smart

library/src/main/java/com/pokegoapi/main/RequestHandler.java
```diff
@@ -274,68 +198,67 @@ private Long getRequestId() {
...
+				// Ignore
 			}
-			if (workQueue.isEmpty()) {
+
+			if (request == null) {
 				continue;
 			}
 
-			workQueue.drainTo(requests);
+			ArrayList<InternalServerRequest> serverRequests = new ArrayList<>();
```
pvanassen: Why ArrayList on the left hand, why no size given to the ArrayList on the right hand, since the size is known

library/src/main/java/com/pokegoapi/main/AsyncServerRequest.java
```diff
@@ -15,60 +15,93 @@
...
-	@Getter
-	private final long id = System.nanoTime();
+public class AsyncServerRequest<T extends GeneratedMessage, K> {
 	@Getter
 	private final RequestType type;
 	@Getter
 	private final Request request;
 	@Getter
-	private final boolean requireCommonRequest;
+	final ArrayList<InternalServerRequest> boundedRequests = new ArrayList<>();
```
pvanassen: Any reason for the left hand ArrayList? Why an array list here, would LinkedList make much more sense?

library/src/main/java/com/pokegoapi/api/player/PlayerProfile.java
```diff
@@ -77,55 +77,91 @@
...
-		} catch (InvalidProtocolBufferException e) {
-			throw new RemoteServerException(e);
-		}
+		new AsyncServerRequest(RequestType.GET_PLAYER, getPlayerReqMsg,
+				new PokeAFunc<GetPlayerResponse, Void>() {
+					@Override
+					public Void exec(GetPlayerResponse response) {
+						parseData(response.getPlayerData());
+
+						ArrayList<TutorialStateOuterClass.TutorialState> tutorialStates =
```
pvanassen: Left hand ArrayList, any reason for this?

library/src/main/java/com/pokegoapi/api/player/PlayerProfile.java
```diff
@@ -77,55 +77,91 @@
...
+							claimCodeName();
+							return null;
+						}
+
+						if (!tutorialStates.contains(
+								TutorialStateOuterClass.TutorialState.FIRST_TIME_EXPERIENCE_COMPLETE)) {
+							firstTimeExperienceComplete();
+						}
+
+						return null;
```
pvanassen: Inner thing always returns null, so why return null in ever if, and not use an if/else?

library/src/main/java/com/pokegoapi/api/map/pokemon/CatchablePokemon.java
```diff
@@ -386,222 +214,148 @@ public CatchResult catchPokemon(double normalizedHitPosition,
...
 					}
-
-					// escapes
-					if (response.getStatus() == CatchStatus.CATCH_ESCAPE) {
-						api.getInventories().updateInventories();
+					//we can retry, but to prevent flagging it's necessary to sleep for a while between each retry
+					try {
+						Thread.sleep(ThreadLocalRandom.current().nextInt(1352, 1976));
+					} catch (InterruptedException e) {
+						e.printStackTrace();
```
pvanassen: Please handle InterruptedExceptions, they are important!

library/src/main/java/com/pokegoapi/api/inventory/PokeBank.java
```diff
@@ -63,12 +55,13 @@ public boolean test(Pokemon testPokemon) {
 	 * @return the pokemon by pokemon id
 	 */
 	public List<Pokemon> getPokemonByPokemonId(final PokemonIdOuterClass.PokemonId id) {
-		return Stream.of(pokemons).filter(new Predicate<Pokemon>() {
-			@Override
-			public boolean test(Pokemon pokemon) {
-				return pokemon.getPokemonId().equals(id);
+		List<Pokemon> ret = new ArrayList<>();
```
pvanassen: Why was this changed, old code did exactly the same it seems

library/src/main/java/com/pokegoapi/api/inventory/CandyJar.java
```diff
@@ -84,10 +63,11 @@ public void removeCandy(PokemonFamilyId family, int amount) {
...
 	 */
 	public int getCandies(PokemonFamilyId family) {
-		if (candies.containsKey(family)) {
-			return this.candies.get(family);
-		} else {
-			return 0;
-		}
+		Integer val = candies.putIfAbsent(family, 0);
+
+		if (val == null)
```
pvanassen: With a putIfAbsent val will never be null, actually, val should be an int. This code can be a one liner

library/src/main/java/com/pokegoapi/api/inventory/CandyJar.java
```diff
@@ -66,14 +50,9 @@ public void addCandy(PokemonFamilyId family, int amount) {
...
 	public void removeCandy(PokemonFamilyId family, int amount) {
-		if (candies.containsKey(family)) {
-			if (candies.get(family) - amount < 0) {
-				candies.put(family, 0);
-			} else {
-				candies.put(family, candies.get(family) - amount);
-			}
-		} else {
-			candies.put(family, 0);
+		Integer current = candies.putIfAbsent(family, amount);
```
pvanassen: Can be int, can never be null

library/src/main/java/com/pokegoapi/api/gym/Battle.java
```diff
@@ -108,10 +102,10 @@ public Result start() throws LoginFailedException, RemoteServerException {
 			gymIndex.add(action.getTargetIndex());
 		}
 
-		return battleResponse.getResult();
+		return battleResponse.getResult();*/
+		return null;
```
pvanassen: What should this do?

library/src/main/java/com/pokegoapi/api/gym/Battle.java
```diff
@@ -198,7 +192,8 @@ private AttackGymResponse sendBlankAction() throws LoginFailedException, RemoteS
 			return AttackGymResponse.parseFrom(serverRequest.getData());
 		} catch (InvalidProtocolBufferException e) {
 			throw new RemoteServerException();
-		}
+		}*/
+		return null;
```
pvanassen: Everything returns null, why having it?

library/src/main/java/com/pokegoapi/api/gym/Battle.java
```diff
@@ -242,8 +236,8 @@ private AttackGymResponse doActions(List<BattleAction> actions) throws LoginFail
 			return response;
 		} catch (InvalidProtocolBufferException e) {
 			throw new RemoteServerException();
-		}
+		}*/
 
+		return null;
```
pvanassen: Again null

library/src/main/java/com/pokegoapi/main/CommonRequest.java
```diff
@@ -166,6 +136,8 @@ public static void parse(PokemonGo api, RequestType requestType, ByteString data
 				case DOWNLOAD_SETTINGS:
 					api.getSettings().updateSettings(DownloadSettingsResponse.parseFrom(data));
 					break;
+				case CHECK_AWARDED_BADGES:
+				//	api.getPlayerProfile().equipBadge(CheckAwardedBadgesResponse.parseFrom(data));
```
pvanassen: Nothing to do here? Why have the case?

library/src/main/java/com/pokegoapi/main/InternalServerRequest.java
```diff
@@ -1,33 +1,18 @@
-/*
```
pvanassen: Removal of license

library/src/main/java/com/pokegoapi/util/PokeCallback.java
```diff
@@ -0,0 +1,31 @@
...
+
+import lombok.Getter;
+
+/**
+ * Created by rama on 31/08/16.
+ */
+public abstract class PokeCallback<T> {
+
+	@Getter
+	private Throwable error;
```
pvanassen: Only Getter, why not final?

library/src/main/java/com/pokegoapi/util/PokeCallback.java
```diff
@@ -0,0 +1,31 @@
...
+	@Getter
+	private T result;
```
pvanassen: Only Getter, why not final?

sample/src/main/java/com/pokegoapi/examples/CatchPokemonAtAreaExample.java
```diff
@@ -33,60 +33,81 @@
...
+							// if encounter was succesful, catch
+							if (encResult.wasSuccessful()) {
+								System.out.println("Encounted:" + cp.getPokemonId());
+								AsyncCatchOptions options = new AsyncCatchOptions(go);
+								options.setRazzBerries(3);
+
+								cp.catchPokemon(options, new PokeCallback<CatchResult>() {
+									@Override
+									public void onResponse(CatchResult result) {
+										System.out.println("CatchResult: " + result.getStatus());
```
pvanassen: I love nesting anonymous inner classes into anonymous inner classes into anonymous inner classes. Absolute marvel for performance.

library/src/main/java/com/pokegoapi/util/PokeCallback.java
ramarro123: because its bot final :) its set by fire method and propagate the event...

pvanassen:
Did a short code review. InterruptedException is never handled and just ignored, making it very hard to stop a thread when it's sleeping. 

Any errors thrown from the callback will cause other CommonRequest not to be handled. 

And more on a general note, the code changes contain a lot of bad practices like using concrete implementations instead of the interface. 

And the examples make a great demonstration why you do not want to use callbacks. Nested 3 layers deep, creating anonymous inner classes inside loops, that is disastrous for performance on any device with limited capabilities.

ramarro123:
i will take a look to all your note later, tnx!

what do you think of making the pokecallback an observable as well as a callback? since its returned in each call you can use it with observwble pattern as well, even without javarx

library/src/main/java/com/pokegoapi/api/gym/Battle.java
ramarro123: as said on previous post, i was not able to kaintain battle code. it was never working on my side, but feel free to contrib if you want

library/src/main/java/com/pokegoapi/api/inventory/CandyJar.java
ramarro123: putifabsent return null if prev value is null?

pvanassen:
I don't see how this would work.

ramarro123:
well pokecall observable and fire update on a list of observer wheb fire its called

pvanassen:
I don't see how changing the existing object will make it work in a similar way like RxJava or any other observable framework. I don't think it can be done without recreating your own Observable framework, which is a waste of time

iGio90:
This is the best way to make the things works and works good. If it's not the "easiest" for the final user, it's not my/our problem. The final users could documents himself to understand how to handle threading/multithreading properly with callbacks. Blocking calls and observables are the easiest way to develop and to get flagged as well. I fully reworked half of my app to use this new impl without problem. If I did it, anyone can. Feel free to keep the 0.4.1 with it's tons of problems then 👍

pvanassen:
Who are you coding for?

iGio90:
PokeMesh

iGio90:
im not coding for. It's my app

iGio90:
that actually count 60.000 users that haven't got a single problem with callbacks

pvanassen:
I code for people using the API. Dismissing any argument by calling the people using your code stupid is not getting you anywhere. 

Good luck.

iGio90:
Thanks :) good luck!

ramarro123:
lol easy guys :) i still think thst there is room for improvement and thats why a community shiuld work together. if i remember well it was asked to contrib actively to this pr isnt it?

feel free to join if you want. 

i will in any case check all the issue that you have pointed out and see if there is room for improvement. 

are you sure of the putifabsend? still cant open javadoc... seems that return jull if the item it's not present.

as for the callback, probably will be hard to change them, but it will be very easy to make them optionally blocking if you think that it will help

Paul776:
@iGio90 You post clearly states that you have very few experience with big-scale/multi-threaded projects. (**which is NOT a problem who cares? this is not a biggest cock competition**)

"If it's not the "easiest" for the final user, it's not my/our problem"
Are you serious? damn..

I had a look at your APK (latest version 6). It's a total mess as I was expecting. Synchronised everywhere. not one single ThreadPool nor java concurrent package utility. using handlers like it was free candies.  

I looked at your "concurrent" scanning it is almost only sequential because of all your badly used synchronised method...

**I do really appreciate the fact that you contribute to the java-api and I can't thank you enough, but don't design it to fit your badly designed android app.** and try understand comments from other fellow (more experience?) java developers which have eaten tons of multi-threaded shit in their life

@iGio90 the callback design pattern is a good pattern just NOT fitted for this scenario. You cannot just throw design pattern here and there jus because they are "cool"

iGio90:
Its just reworked to fit the latest changes on the api. Still lot of rework needs to be done to fit it. Im not saying that is perfect nor that im that awesome developer in terms of multithreading and threadpools. What i'm saying is that im not seeing any active contribution. What i saw are just repetitive post that says "blocking is the way". And yeah, I confirm that we are not here to make the things "easiest for developers" instead, we are here to make them works in the proper way

pvanassen:
'feel free to join if you want'.
I only got, we do it our way and are not going to look at or listen to anyone else. 

**I also do really appreciate the fact that you contribute to the java-api and I can't thank you enough, but this is a very very bad design.**

I'm off to my own happy corner with my own fork.

iGio90:
I wonder what you can see by reversing the application with it's pseudo code instead of taking a look at the source code. Synchronized everywhere is necessary to hold a proper multiaccount scan. Im not interested anyway in reversing other apps to check how them works. My badly design app have an avarage of 60k users, your good design one can't probably reach 10k. So please, let's speak less of how my app is designed if you can't handle or you are asking how to properly handle errors with callbacks. 

@Paul776 no one said that we are not open for contribution. IMO keep saying that callbacks are not a proper way is stating that you didn't understood a thing on how the official client is working

iGio90:
I.E the actual api, without this PR, is firing request in a random orders, and it's not bounding common requests within the same envelope

iGio90:
Do you know how the request envelope is built at least?

iGio90:
You are reversing my app to check its "badly design code" and no one noticed that this library had the user-agent "OkHttpClient" before i pushed the commit? You are so good to inspect other code and you cant open the request handler to see this:

builder.setMsSinceLastLocationfix(989);

I wonder why you suggest something then

iGio90:
And don't try to say that im building this API just for my "badly" design app when you got ur own impl on ur github that is designed for your app, and you didn't submit a single PR to this repo

iGio90:
@Paul776 it's ok.. GG!

Paul776:
@iGio90 
I'm done here. I'd suggest you to take step back, try to understand the pros and cons of continuing with this design.  I'm maybe not contributing by pushing code but i'm by giving you my input on app architecture.

Issue #727, opened 2016-08-30
Profile not updating?


bbb0x:
I created a simple test bot; placed him between 2 pokestops and all he does is checking every min if he can loot them, if yes it uses them.

I print the player XP before and after the loot and it always shows the same.
Only restarting gets me the new value.

Here the 2 functions I call before and after (updateplayer is called first) the loot:

```
    private void UpdatePlayer() throws RemoteServerException, LoginFailedException
    {
        if(profile == null){
        profile = go.getPlayerProfile();
        stats = profile.getStats();
        }
        else
        profile.updateProfile();
    }

    private void PrintStats()
    {

        int level = stats.getLevel();
        long experience = stats.getExperience();
        long toNext = stats.getNextLevelXp();


        MCLog.Log("Player level: "+level);
        MCLog.Log("XP: "+experience);
        MCLog.Log("To next Level: "+toNext);
    }

```

Both times, XP prints the same value. NextLevelXP too but I assume the meaning of that function is that it shows the entire XP a level needs, so it's fine.

FabianTerhorst:
you have to call api.getPlayerProfile().updateProfile()

FabianTerhorst:
When updateProfile doesn´t work this would be strange.

bbb0x:
I tested 2 things but both didn't work:

```

        go.getPlayerProfile().updateProfile();
        profile = go.getPlayerProfile();
        stats = profile.getStats();

```

and

```
        profile = go.getPlayerProfile();
        profile.updateProfile();
        stats = profile.getStats();
```

Not sure, I'm more a C# guy so I'm new on Java anyway. Are these functions maybe maybe async that it jumps to my PrintStats before the update is finished?

bbb0x:
Being busy with the issue for the last couple of hours and the solution was very weird.

```
`go.getInventories().updateInventories();`
```

This line fixed it, but shouldn't it be PlayerProfile and not Inventories that updates the stats? Updating the inventories in my update isn't wrong; but I wasn't at that point in my development yet to manage the inventories.

FabianTerhorst:
The stats are inside the inventory api response. Thats the reason for this.

PR #737, opened 2016-08-31
Inventories and Settings not loaded at initialization.

gdespres:
**Fixed issue:**
 #727

**Changes made:**
index of Inventories and Settings request are wrong

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/PokemonGo.java
+++ b/library/src/main/java/com/pokegoapi/api/PokemonGo.java
@@ -210,8 +210,8 @@ public class PokemonGo {
 
 		getRequestHandler().sendServerRequests(requests);
 		try {
-			inventories.updateInventories(GetInventoryResponse.parseFrom(requests[2].getData()));
-			settings.updateSettings(DownloadSettingsResponse.parseFrom(requests[4].getData()));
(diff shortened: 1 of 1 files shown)


FabianTerhorst:
@iGio90 review please

FabianTerhorst:
why this is fixing #727? The inventory api is providing the stats. So when you want now stats you have to update the inventory. https://github.com/Grover-c13/PokeGOAPI-Java/blob/Development/library/src/main/java/com/pokegoapi/api/inventory/Inventories.java#L155

iGio90:
@FabianTerhorst yeah, the index are changed when I added the check challenge at the first position. However, my rework on PR will get rid of all of this things and parse each response from the common request class

iGio90:
Maybe it won't fox #727 but the fix is right in any case

FabianTerhorst:
ok then, but #727 isn´t an issue.

gdespres:
Maybe it won't fix #727, but I think there is a link with the issue because Inventories was not updated at login().

call manualy go.getInventories().updateInventories() resolve it.

gdespres:
now inventories (and settings) are updated at login() and manualy call go.getInventories().updateInventories() is not needed

FabianTerhorst:
its needed when you catched a pokemon.

FabianTerhorst:
So that the xp is changing.

Issue #739, opened 2016-08-31
Map#getMapObjectsAsync uses cache incorrectly


Gegy:
When Map#getMapObjectsAsync checks whether it should use the cache or not, it checks if cachedCatchable has any entries instead of checking if cachedMapObjects has any entries.

https://github.com/Grover-c13/PokeGOAPI-Java/blob/Development/library/src/main/java/com/pokegoapi/api/map/Map.java#L335

FabianTerhorst:
@LoungeKatt please fix your issue.

FabianTerhorst:
https://github.com/Grover-c13/PokeGOAPI-Java/commit/910078b8e09528b38bae65e17b4304f0be7e54ed#diff-e356103bee63c77e4131618df05d6ff0R335

gdespres:
Why verify cachedCatchable size ? PR #743 fix the issue.

AbandonedCart:
@FabianTerhorst you verified and approved the code (via slack), claiming to understand it at the time. 

The explanation was provided at https://github.com/Grover-c13/PokeGOAPI-Java/pull/695/commits/910078b8e09528b38bae65e17b4304f0be7e54ed

On a side note, it is always good for people to look at a commit when it includes code that seems "misplaced" to verify there isn't some sort of explanation for it. This one included a note that allows following the code back to where the objects are filled to find that the object in question is a part of it.

LoungeKatt:
https://github.com/Grover-c13/PokeGOAPI-Java/pull/695/commits/910078b8e09528b38bae65e17b4304f0be7e54ed

PR #743, opened 2016-09-02
Bad Copy / Paste

gdespres:
**Changes made:**
Delete the cacheCatchable size verification to use mapObjetsCache

*\* Fix issue **
#739

No need to verifiy the size of cachedCatchable to use the cachedMapObjects.
When you use the getMapObjects().getPokestops() (so cachedCatchable == null) method first and after getCatchablePokemon(), the cachedMapObjects should be used and it is not the case.

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/map/Map.java
+++ b/library/src/main/java/com/pokegoapi/api/map/Map.java
@@ -332,7 +332,7 @@ public class Map {
 	 */
 	public Observable<MapObjects> getMapObjectsAsync(List<Long> cellIds) {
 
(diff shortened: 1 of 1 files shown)


AbandonedCart:
This wasn't a "bad copy/paste"
It was intended to provide a method to force refreshing the cache when it was cleared by external means. By removing this, you can end up clearing the cache by changing location but it will continue to attempt loading the empty cache until the boolean is reset elsewhere.

AbandonedCart:
I'll merge this and then add the proper (lengthy) fix to still verify there are items. As it stands this will still block filling in items when the cache is empty but has been reset.

Issue #758, opened 2016-09-06
CatchOptions.useRazzberries(boolean) not working as expected


bbb0x:
I had the problem that when I catched a Pokemon, it threw an exception.
Analyzing the problem I saw that it occured when I had no razzberies even through I set after a check if I still have some razzberies:

`options.useRazzberries(false);`

I had to fix it by adding this line just after it:

`options.maxRazzberries(0);`

The first function should set this internal already IMO.

gsisso:
thats what I said at #634 but @LoungeKatt said it will break it somewhere else.

AbandonedCart:
It does set it internally at https://github.com/Grover-c13/PokeGOAPI-Java/blob/Development/library/src/main/java/com/pokegoapi/api/settings/CatchOptions.java#L73
Part of the issue is https://github.com/Grover-c13/PokeGOAPI-Java/blob/Development/library/src/main/java/com/pokegoapi/api/settings/CatchOptions.java#L170 not setting the value before https://github.com/Grover-c13/PokeGOAPI-Java/blob/Development/library/src/main/java/com/pokegoapi/api/settings/CatchOptions.java#L179 
The explanation should also clarify that it is meant as an override for using a single berry without actually setting a max, not that it is a prerequisite or blanket enabler.

PR #761, opened 2016-09-07
[Issue #758] CatchOptions#useRazzberries(boolean) not working

yutamago:
**Fixed issue:** #758 

**Changes made:**
- removed unnecessary use of 1 razzberry prior to actual use of n razzberries
- fixed CatchOptions#getRazzberries()
- fixed bug that used more razzberries than available
- added two Log.d's that notify you about how many razzberries you're about to use on the encountered Pokémon, and how many razzberries were actually used.

I ran this with a quickly made bot for a few hours without any errors, catching about 3 Pokemon per minute.
Feel free to test this in a productive environment. I don't have any at this point.
Maybe @LoungeKatt ?

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/map/pokemon/CatchablePokemon.java
+++ b/library/src/main/java/com/pokegoapi/api/map/pokemon/CatchablePokemon.java
@@ -438,8 +438,16 @@ public class CatchablePokemon implements MapPoint {
 		Item razberriesInventory = api.getInventories().getItemBag().getItem(ItemId.ITEM_RAZZ_BERRY);
 		int razberries = 0;
 		int numThrows = 0;
-		Log.d("catchPokemon", String.format("attempt to catch pokemon using max %d razzberries. %d in inventory",
-				razberriesLimit, razberriesInventory.getCount()));
+		Log.d("catchPokemon",
+				String.format("attempt to catch pokemon%s.",
+						razberriesLimit > 0
+								? String.format(" using max %d razzberrys. %d in inventory",
+								razberriesLimit, razberriesInventory.getCount())
+								: razberriesLimit == -1
+								? String.format(" using all %d razzberrys in inventory",
+								razberriesInventory.getCount())
+								: " using no razzberries"
+				));
 
 		CatchResult result;
 		do {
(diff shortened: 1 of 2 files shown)


AbandonedCart:
This looks like a regression that re-implements forcing the use of razzberries by default.

AbandonedCart:
https://github.com/Grover-c13/PokeGOAPI-Java/pull/766 will fix the razzberries without changing the functionality to force a default of 1 or require setting both the use to true and the max, since not setting use at all will default to false and invalidate any max value set using the code here.

yutamago:
My code is equivalent to the previous code, but fixing said issue.

**Previously**
Defaults: `useRazzBerries = false; maxRazzBerries = 0;`
getRazzBerries(): `return useRazzBerries && maxRazzBerries == 0 ? 1 : maxRazzBerries;`

**Changed**
Defaults: `useRazzBerries = false; maxRazzBerries = 1;`
getRazzBerries(): `return !useRazzBerries ? 0 : maxRazzBerries;`

### Cases

Case 1: We create a **default** instance of CatchOptions
Previously: getRazzBerries() == 0
Changed: getRazzBerries() == 0

Case 2: We then call **useRazzBerries(true)**
Previously: getRazzBerries() == 1
Changed: getRazzBerries() == 1

Case 3: We then call **maxRazzBerries(10)**
Previously: getRazzBerries() == 10
Changed: getRazzBerries() == 10

Case 4: We then call **useRazzBerries(false)** <--- This is what the issue is about
Previously: getRazzBerries() == 10 // **unexpected**
Changed: getRazzBerries() == 0     // expected

Case 5: We then call **maxRazzBerries(-1)**
Previously: getRazzBerries() == -1 // **unexpected**
Changed: getRazzBerries() == 0    // expected

Case 6: We then call **useRazzBerries(true)**
Previously: getRazzBerries() == -1
Changed: getRazzBerries() == -1

I don't think this is a misunderstanding by the issue creator. The Javadocs say following:
useRazzBerries(boolean): `Enable or disable the use of razzberries`
getRazzBerries(): `@return the number [of razzberries] to use`

AbandonedCart:
TL;DR it is a misunderstanding in how it was intended. As stated, you should never be required to set two variables to do a single function. I get that you are trying to make the use variable a universal use or not use and then max the actual number, but the intention was that use would be a shortcut to use them (period) and max would be how to set a certain number, but none would be used by default.
In this version 1 is used by default. 1 is not 0, no matter how you explain it.

yutamago:
The previous default was 1 as well if you look at getRazzBerries().
As soon as you useRazzBerries(true) it would return 1. It would overwrite the default value of maxRazzBerries. If you would take these 15 seconds to fly over my comment, you would see for yourself.

AbandonedCart:
https://github.com/Grover-c13/PokeGOAPI-Java/pull/761/commits/e369d22c5e17db5a9a2d9df34a773f900e51c711#diff-27339d567cdc353929a6e9d986dc73dbR72 is a default. 
Setting something is not a default. What you are saying is by setting something, it changed how it worked. That is the entire purpose of setting things.

PR #766, opened 2016-09-08
Remove legacy implementation for single razzberries

AbandonedCart:
### Prerequisites (Remove this section if you want)

Make sure you...
- Follow the [contribution guidelines](CONTRIBUTING.md)
- Follow the code style (run `./gradlew checkstyleMain`)
- Submit this PR against the `Development` branch.

**Fixed issue:** [Reference the issue number here, or remove if not a fix]

**Changes made:**
- List your changes here
- Change 2...

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/map/pokemon/CatchablePokemon.java
+++ b/library/src/main/java/com/pokegoapi/api/map/pokemon/CatchablePokemon.java
@@ -252,13 +252,7 @@ public class CatchablePokemon implements MapPoint {
 	 */
 	public CatchResult catchPokemon(CatchOptions options) throws LoginFailedException,
 			RemoteServerException, NoSuchItemException {
-		if (options != null) {
-			if (options.getRazzberries() == 1) {
-				useItem(ItemId.ITEM_RAZZ_BERRY);
-				options.useRazzberries(false);
-				options.maxRazzberries(-1);
-			}
-		} else {
+		if (options == null) {
 			options = new CatchOptions(api);
 		}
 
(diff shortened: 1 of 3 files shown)


yutamago:
But the issue still exists if you let this line unchanged: 
`return useRazzBerries && maxRazzBerries == 0 ? 1 : maxRazzBerries;`

**Example**: You use one instance of CatchOptions throughout your project and at some point you want to deactivate the use of razzberries. Instinctively you just call useRazzBerries(false).

Let's say maxRazzBerries was 10 before:
1. `return useRazzBerries && maxRazzBerries == 0 ? 1 : maxRazzBerries;`
2. `return False && 10 == 0 ? 1 : 10;`
3. `return False ? 1 : 10;`
4. `return 10;`

10 razzberries will be used.

In my implementation it will prioritize useRazzBerries over maxRazzBerries.
The same Example will use exactly 0 razzberries.

Or did I miss something?

LoungeKatt:
You are missing that the intended functionality is what you are trying to remove. I guess I need to change the variable name so it will make sense. Originally, maxRazzBerries was the only variable but that confused a few people. useRazzBerries was added so they had a way to use just one without explicitly setting 1. Some people like the true / false without having to think in numbers.

AbandonedCart:
This is explained in the other PR. It appears the variable name and possibly description have confused a few people. It will be adjusted.

AbandonedCart:
@dojo1017 Appending the fixes for checking the inventory from https://github.com/Grover-c13/PokeGOAPI-Java/pull/761 would still be a good idea. I tried to keep the major code changes somewhat similar so they could be merged over this.

PR #770, opened 2016-09-08
Check razzberries in inventory before use and update after

yutamago:
**Changes made:**
We now check if razzberries are available before using them.
The inventory will update after use.

@LoungeKatt is that how you meant it? I made it a bit more readable.


AbandonedCart:
Exactly. Thanks!

PR #790, opened 2016-11-09
Update Hash function and protos

Gegy:
This PR updates the hash function and protocol to enable GetMapObjects to return valid results once again.

Updated Java Hash function from: https://github.com/POGO-Java/UK6 and https://github.com/br0x/NiaHash


DaRealGunko:
Make new version for gradle pls.

PR #795, opened 2016-11-12
Updated for 0.45 new pokemons

PokeBotPub:
Updated for 0.45 new pokemons


PokeBotPub:
You are need update PokemonId.proto
URL : https://github.com/AeonLucid/POGOProtos/blob/master/src/POGOProtos/Enums/PokemonId.proto

Gegy:
Pokemon are now loaded from the server, thanks for the PR. :)

PR #796, opened 2016-11-12
Fix empty map objects when not on altitude 0

Gegy:
This Pull Request fixes getMapObjects returning no results when the altitude was not 0. The hashes were supposed to use the accuracy and not the altitude. I have also added accuracy fields to the main API class.


iGio90:
This is wrong. Its not using horizontal accuration. You are always hashing with third param as -1.
Params for the hashes are lat lon and field9 from requestenvelope. Field9 on the envelope is accuracy

PR #820, opened 2016-12-08
General improvements and fixes

Gegy:
This PR improves common request handling, adds medal and buddy support, fixes some problems with captchas, and adds path utilities to easily move the player between two points at the desired speed.

To go with the path utils, a loot pokestop example was added, which walks at 15KMPH to a near pokestop and loots it.
This also adds some new listeners for buddies and medals.

### DIFF
--- a/library/src/main/java/com/pokegoapi/api/map/Point.java
+++ b/library/src/main/java/com/pokegoapi/api/map/Point.java
@@ -39,4 +39,9 @@ public class Point implements MapPoint {
 		this.latitude = spawnpoint.getLatitude();
 		this.longitude = spawnpoint.getLongitude();
 	}
+
+	@Override
+	public String toString() {
+		return this.latitude + ", " + this.longitude;
+	}
 }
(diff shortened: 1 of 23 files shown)


library/src/main/java/com/pokegoapi/api/map/Point.java
```diff
@@ -39,4 +39,9 @@ public Point(SpawnPointOuterClass.SpawnPoint spawnpoint) {
 		this.latitude = spawnpoint.getLatitude();
 		this.longitude = spawnpoint.getLongitude();
 	}
+
+	@Override
+	public String toString() {
+		return this.latitude + ", " + this.longitude;
```
PokeAlert, line 45: StringBuilder

library/src/main/java/com/pokegoapi/api/pokemon/Buddy.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+package com.pokegoapi.api.pokemon;
+
+import POGOProtos.Data.BuddyPokemonOuterClass.BuddyPokemon;
+import com.pokegoapi.api.PokemonGo;
+
+public class Buddy {
+	private final PokemonGo api;
+	private final BuddyPokemon proto;
```
PokeAlert, line 23: proto stored in memory? that cant be good memory wise

PR #832, opened 2016-12-27
Initial hashing service

Pat-ji:
Be sure to call PokemonGo#setHashKey(KEY) directly after creating your
instance.

*MapObjects might need an update to work
*Currently uses the HTTP endpoint, change BosslandHash#BASE if you wish to use a different endpoint
*Might need better failed request handling
*Feel free to improve upon this

### DIFF
--- a/library/src/main/java/com/pokegoapi/util/BosslandHash.java
+++ b/library/src/main/java/com/pokegoapi/util/BosslandHash.java
@@ -1,0 +1,125 @@
+package com.pokegoapi.util;
+
+import okhttp3.*;
+import com.pokegoapi.api.PokemonGo;
+import lombok.Getter;
+import lombok.Setter;
+
+import java.io.IOException;
+import java.lang.IllegalAccessException;
+import java.lang.IllegalStateException;
+import java.lang.NullPointerException;
+import java.lang.System;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Base64;
+
+public class BosslandHash {
+    private static final String BASE = "http://pokehash.buddyauth.com/";
+    private static final String VERSION_END = "api/hash/versions";
+
+    private static String version;
+
+    public static class HashResponseContent {
+        @Getter private long locationAuthHash;
+        @Getter private long locationHash;
+        @Getter private List<Long> requestHashes;
+
+        public HashResponseContent(String response) {
+            response = response.replaceAll("\\D+", "|").substring(1);
+            String[] split = response.split("\\|");
+
+            locationAuthHash = Long.parseLong(split[0]);
+            locationHash = Long.parseLong(split[1]);
+
+            requestHashes = new ArrayList<>();
+            for (int i = 2; i < split.length; i++)
+                requestHashes.add(Long.parseLong(split[i]));
+        }
+
+    }
+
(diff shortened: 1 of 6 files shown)


library/src/main/java/com/pokegoapi/util/BosslandHash.java
```diff
@@ -0,0 +1,125 @@
...
+        public String json() {
+            Base64.Encoder encoder = Base64.getEncoder();
+            String auth = encoder.encodeToString(authTicket);
+            String session = encoder.encodeToString(sessionData);
+            String request = "";
+            for (byte[] array : requests)
+                request += "\"" + encoder.encodeToString(array) + "\",";
+            request = request.substring(0, request.length() - 1);
+
+            return "{"
```
PokeAlert, line 64: Why are you not using the already imported Moshi

library/src/main/java/com/pokegoapi/util/BosslandHash.java
```diff
@@ -0,0 +1,125 @@
...
+            case 200: // OK
+                String result = response.body().string();
+                response.close();
+                return new HashResponseContent(result);
+            case 400: // Bad request
+                throw new IllegalStateException("Bad request to hashing server, request=" + content.json());
+            case 401: // Unauthorized
+                throw new IllegalStateException("Unauthorized to request hash values");
+            case 429: // Limit exceeded
+                throw new IllegalStateException("Exceeded hashing limit");
```
PokeAlert, line 119: Not retrying? request lost?

library/src/main/java/com/pokegoapi/util/Crypto.java
```diff
@@ -38,12 +38,8 @@
 
 	private static byte makeIntegrityByte(Rand rand) {
 		rand.state = (0x41C64E6D * rand.state) + 0x3039;
-		long shiftedRand = rand.state >> 16;
-		byte lastbyte = Long.valueOf(shiftedRand).byteValue();
-
-		byte v74 = (byte) ((lastbyte ^ 0x0C) & lastbyte);
-		byte v75 = (byte) (((~v74 & 0x67) | (v74 & 0x98)) ^ 0x6F | (v74 & 8));
-		return v75;
+		byte b = Long.valueOf(rand.state >> 16).byteValue();
```
PokeAlert, line 41: I did a test I still get response by removing this
only returning return (byte) (rand.state.byteValue() & 0xE3 | 0x10);
is enough. Can someone confirm?

library/src/main/java/com/pokegoapi/util/BosslandHash.java
Pat-ji: Didn't have the internet capacity to check out the imported libraries and how they work. I can add it when I get back home

library/src/main/java/com/pokegoapi/util/BosslandHash.java
Pat-ji: It probably should, this was just a raw convert to get it working. If noone else does it, I'll gladly improve on it when I get back home

Gegy:
#833 has support for both this, and the legacy hasher.

PR #860, opened 2017-02-12
Fix typo in login example in README

pgier:


Gegy:
We should probably change the README example further to account for other changes that have occurred since it was last edited.

AbandonedCart:
The last edit was Sep 6, 2016 for grammar. That was before the capcha and a whole mess of security changes, so it is really a matter of when, not if, it is updated. Luckily the game isn't as popular anymore or the outdated readme would have led to a lot of unnecessary issue reports.

AbandonedCart (approved):

PR #865, opened 2017-02-15
Rework request system, request accuracy & 0.57.4

Gegy:
This PR reworks the request system to allow envelope construction to be done by the API user, and not the API. Instead of sending all queued requests every second, the user will provide all requests to send in a single envelope. This has allowed for the initial blank request to be cleanly implemented.


Gegy:
SESSION_INVALIDATED is being received from the initial blank request - working on a fix.

PR #873, opened 2017-02-19
[In progress]PokeGOAPI-Java version 2.0 Providers Framework rewrite

s7092910:
A complete rewrite of the Java version of the Pokemon Go Api. The goal of this rewrite is to bring better maintainability, flexibility, ease of use and modularity.

 To achieve those goals, the rewrite will be mainly using the Provider design pattern, where there are classes that handles the interactions between the library user and the implementations of the service provider interfaces(SPI). A provider class can be created by the library user to implement their own implementations of the SPIs and supply the provider class to the class handling the interactions.

Along with the provider design pattern, modularity will be introduced into the PokemonGoClient. For instance, if a user does not care about knowing the Pokemon Go account's inventory, the user can select to not add the inventory related classes to the PokemonGoClient.

Along with these changes, there will be the removal of software bloat and libraries that are not required due to these rewrites.

Overall this will be a long task that will hopefully improve the library/api.

Any input is welcomed and people wanting to help out are free to do so.


AdrienNguyenWorldline:
I think this should not even be a fork but a standalon project.

vamidi:
Did pokemon go makers changed the code or is it still the same and is this project still ongoing?
I hope you will succeed in this project 👍

PR #880, opened 2017-03-01
Refactoring about MapPoint and Pokemon

DrumRobot:
* Create interface **IPokemon** and refactoring classes about Pokemon to implement **IPokemon**
* Refactoring **Pokestop** class to implement **MapPoint**
* Using **MapPoint** instead of **Point** if possible
* Replace **CatchablePokemon::encounterPokemon** returning types to **Encounter** class
* Prevent **NoSuchItemException** from **PokemonDetails::getMaxCp**


s7092910:
A few things, you are missing a description of this pull request. Second, it seems you are missing a few changed classes in your commit. Namely Encounter and the new Encounter classes. Thirdly, smaller commits are better. For instance your one commit should be at least 3 commits.

Usually it is good for 1 commit to be about one thing or issue or fix. For instance, refactoring MapPoint should be its own commit, etc.

Gegy:
You've also included all the commits from my fork, but removed some important code from it. It seems you've added checkstyle and suppression files, they should not be there.

PR #900, opened 2017-04-08
Fix #899

### LINKED ISSUE #899: encounter.useItem always returns ACTIVE_ITEM_EXISTS
It is not possible to use berries when catching a pokemon.
encounter.useItem always returns ACTIVE_ITEM_EXISTS

You can check this with the CatchPokemonAtEreaExample

I changed line 178 from
encounter.useItem(ItemId.ITEM_RAZZ_BERRY);
to
System.out.println(encounter.useItem(ItemId.ITEM_RAZZ_BERRY).toString());
to get the result.

Gegy:


library/src/main/java/com/pokegoapi/api/inventory/Inventories.java
```diff
@@ -139,6 +139,12 @@ public void updateInventories(GetInventoryResponse response) {
 
 		for (InventoryItemOuterClass.InventoryItem inventoryItem
 				: response.getInventoryDelta().getInventoryItemsList()) {
+			
+			// Remove released Pokemon from bag.
+			if(inventoryItem.getDeletedItem().getPokemonId() != 0) {
```
jabbink, line 144: [ant:checkstyle] [WARN] /home/jasper/Documents/PokemonGo/PokeGOAPI-Java/library/src/main/java/com/pokegoapi/api/inventory/Inventories.java:144:9: WhitespaceAround: 'if' is not followed by whitespace. Empty blocks may only be represented as {} when not part of a multi-block statement (4.1.3) [WhitespaceAround]

Gegy: @RebliNk17

PR #903, opened 2017-04-09
Support for the new 0.61.0 API

RebliNk17:
Added support for the new 0.61.0 api.

Added a Settings listener.
The settings listener should be add as the other listeners in the api.


library/src/main/java/com/pokegoapi/util/hash/pokehash/PokeHashProvider.java
```diff
@@ -39,14 +39,16 @@
...
 	@Getter
 	@Setter
 	private String endpoint = DEFAULT_ENDPOINT;
 
-	private static final int VERSION = 5901;
-	private static final long UNK25 = -3226782243204485589L;
+	private static final int VERSION = 6100;
+	private static final long UNK25 = 1296456256998993698L;
+
+	private static final int[] currentAPIVersion = new int[] {0,61,0};
```
Gegy: `CURRENT_VERSION`?

library/src/main/java/com/pokegoapi/api/settings/Settings.java
```diff
@@ -123,6 +123,23 @@ public void updateSettings(DownloadSettingsResponse response) {
 		if (response.getSettings().hasGpsSettings()) {
 			gpsSettings.update(response.getSettings().getGpsSettings());
 		}
+		// check if new api has been forced by Niantic
+		if(!response.getSettings().getMinimumClientVersion().isEmpty()) {
+			int[] currentVer = api.getHashProvider().getCurrentAPIVersion();
+			String[] curVer = response.getSettings().getMinimumClientVersion().split("\\.");
+			int[] minimumVer = new int[currentVer.length];
+			for(int x = 0; x < currentVer.length; x++)
+				minimumVer[x] = Integer.valueOf(curVer[x]);
```
Gegy: `Integer.valueOf(String)` is for returning an `Integer` object, here you can just use `Integer.parseInt(String)`. You should also handle the `NumberFormatException`.

library/src/main/java/com/pokegoapi/api/settings/Settings.java
```diff
@@ -123,6 +123,23 @@ public void updateSettings(DownloadSettingsResponse response) {
 		if (response.getSettings().hasGpsSettings()) {
 			gpsSettings.update(response.getSettings().getGpsSettings());
 		}
+		// check if new api has been forced by Niantic
+		if(!response.getSettings().getMinimumClientVersion().isEmpty()) {
+			int[] currentVer = api.getHashProvider().getCurrentAPIVersion();
+			String[] curVer = response.getSettings().getMinimumClientVersion().split("\\.");
+			int[] minimumVer = new int[currentVer.length];
+			for(int x = 0; x < currentVer.length; x++)
```
Gegy: Maybe something more descriptive than `x`?

library/src/main/java/com/pokegoapi/api/settings/Settings.java
```diff
@@ -123,6 +123,23 @@ public void updateSettings(DownloadSettingsResponse response) {
 		if (response.getSettings().hasGpsSettings()) {
 			gpsSettings.update(response.getSettings().getGpsSettings());
 		}
+		// check if new api has been forced by Niantic
+		if(!response.getSettings().getMinimumClientVersion().isEmpty()) {
+			int[] currentVer = api.getHashProvider().getCurrentAPIVersion();
+			String[] curVer = response.getSettings().getMinimumClientVersion().split("\\.");
```
Gegy: `curVer` implies that this is the current version of the API, and this conflicts with the above `currentVer`. You then use `minimumVer` for the same values as ints. Maybe use `splitMinimumVer` instead?

Gegy:
Can you update the protos?

PR #929, opened 2017-07-18
0.67.2 Gym display error fix

yearchan:


library/src/main/java/com/pokegoapi/api/gym/Gym.java
```diff
@@ -155,29 +155,28 @@ private GetGymDetailsResponse details() throws RequestFailedException {
 
 	public String getName() throws RequestFailedException {
 		return details().getName();
-	}
+	}	
 
-	public ProtocolStringList getUrlsList() throws RequestFailedException {
-		return details().getUrlsList();
+	public String getUrlsList() throws RequestFailedException {
```
Gegy: Since it is only one url, maybe rename the method to just `getUrl`? The signature is changed anyway when you change the return type.

yearchan: many thx your help and correction. i will update after finished the loot function tmr.

PR #932, opened 2017-07-19
Feed Pokemon

yearchan:


library/src/main/java/com/pokegoapi/api/gym/Gym.java
```diff
@@ -161,6 +166,50 @@ private GymGetInfoResponse details() throws RequestFailedException {
...
+			try {
+				feedPokemon = GymFeedPokemonResponse.parseFrom(serverRequest.getData());
+				if (feedPokemon.getResult().equals(Result.SUCCESS)) {
+					item.setCount(startingQuantity - 1);
+				}
+			} catch (InvalidProtocolBufferException e) {
+				throw new RequestFailedException();
+			}
+		} else {
+			throw new RequestFailedException("Not Enough Item");
```
gegy1000: It may be better to return a response with result `ERROR_NO_BERRIES_LEFT`. Maybe also handle other cases like being on the incorrect team.

Gegy:
We do not want to support gym interaction.

AbandonedCart:
Since when do "we" not support certain parts? The entire purpose of the API is to provide access to every aspect of the game externally.

jabbink:
>The entire purpose of the API is to provide access to every aspect of the game externally.

Except it hasn't and it has turned to shit ever since people decided to add dozens of useless helper functions for e.g. catching Pokemon instead of just plainly exposing all functionality this game has to offer.

Gegy:
By "we" I was referring to the PoGo Dev community. In the context of this PR and project, this does not make sense and closing this PR was not right and I apologize. These matters are something that should be discussed.

jabbink:
imho an api implementation should forward all functionality available in the protocol. Because right now some might say interaction with gyms is unethical. Then where do you draw the line? Apparently catching Pokemon is fine, which still has an influence on regular game play, when Niantic would ever give the game trading.

Gegy:
It could be debated that catching pokemon would have the same advantage to gyms that it would have with trading. Trading could also be considered "unethical", and in the case that this library took the route of not implementing unethical implementations, trading along with gyms would not be implemented.

jabbink:
Unfortunately, when trading would ever become a thing, removing the ability to catch Pokemon is not advisable either, even if unethical, as Niantic apparently bans, by empirical evidence, accounts that just do gmo requests without ever catching and spinning pokestops, so all maps, which I don't see as unethical, are actually also doing a lot with their inventory like catching and upgrading Pokemon, in order to not get banned.

I don't see Niantic ever doing this for gyms, as people living in rural areas will hardly ever encounter a gym, or take it and then can't come back as it takes weeks before someone else attacks the gym.

Gegy:
Yes, removing the ability to catch pokemon would be a bad decision, as that is one of the core parts of the game. What I meant before, is that if this library took the route of not implementing unethical features like gym interaction, instead of pokemon catch not being implemented, trading itself would not be implemented.

AbandonedCart:
The library shouldn't take a stance on anything. It is a library. It provides an API so each developer can make an ethical decision about their own implementation. The code does not automatically walk to the gym, attack it, ensure a victory, and then maintain it. It simply provides an interface and the parameters available to the game. The second you begin categorizing what is and isn't ethical for an API, the entire argument that it is "educational" and "legal" becomes questionable.

